PR #4, opened 2022-03-23
[WIP] Add KDP - Mutate Existing Resources

realshuting:
Signed-off-by: ShutingZhao <shuting@nirmata.com>

### DIFF
--- a/proposals/mutate_existing_resources.md
+++ b/proposals/mutate_existing_resources.md
@@ -1,0 +1,213 @@
+# Kyverno Design Proposal - Mutate Existing Resources
+
+Created: March 21st, 2022
+Author: [Shuting Zhao](https://github.com/realshuting)
+
+
+Table of contents:
+  * [Overview](#overview)
+  * [Proposal](#proposal)
+  * [Examples](#examples)
+  * [Implementation](#implementation)
+    + [Watch Trigger Resources](#watch-trigger-resources)
+      - [1. Admission Webhook](#1-admission-webhook)
+      - [2. Event Watcher](#2-event-watcher)
+    + [Apply Policies](#apply-policies)
+    + [Mutate Existing Resources](#mutate-existing-resources)
+
+<small><i><a href='http://ecotrust-canada.github.io/markdown-toc/'>Table of contents generated with markdown-toc</a></i></small>
+
+
+## Overview
+
+Kyverno webhook controller runs as an admission controller in Kubernetes, and it mutates resources upon admission review. Such mutation is performed based on resources operations (CREATE/UPDATE/DELETE) via webhook.
+
+As of Release 1.6.1, Kyverno has growning supprot for mutating existing resources:
+
+- Mutate target resource which is different from the trigger resource, [#2139](https://github.com/kyverno/kyverno/issues/2139), [#1722](https://github.com/kyverno/kyverno/issues/1722).
+- Mutate trigger resources based on policy update, [#1607](https://github.com/kyverno/kyverno/issues/1607).
+
+*The **trigger** resource is matched in the policy while the **target** resource is the resource to be mutated*.
+
+
+## Proposal
+
+To support mutate existing resources:
+
+- `spec.rules.match` will be used to identify operations on the trigger resources
+    - To *"mutate trigger resources based on policy update"*, the syntax is slightly different in order to simplify policy definition, see example 3 below.
+- A new attribute `spec.rules.mutate.targets` will be introduced to specify a list of target resources to be mutated
+- Kyverno will fetch and mutate the existing resources once the trigger resource changes, the user needs to grant Kyverno permissions to operate on target resources
+- The patches will be added as an annotation to existing resources.
+
+
+
+
+## Examples
+
+1. This exmaple adds the label `foo=bar` to deployment `staging/example-A` on configmap `staging/dictionary` update.
+
+```yaml=
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: "..."
+spec:
+  rules:
+    - name: "..."
+      match:
+        resources:
+          kinds:
+          - ConfigMap
+          names:
+          - dictionary
+          namespaces:
+          - staging
+      mutate:
+        # specify target resources to be mutated
+        # 
+        # if this tag is specificied, Kyverno will 
+        # automatically perform post mutation, there's 
+        # no need to enable this feature additionally
+        targets:
+          # starts with GVK, can be extended to suppport
+          # labels, annotations, etc
+        - kinds:
+          - Deployment
+          names:
+          - example-A
+          namespaces:
+          - staging
+        patchStrategicMerge:
+          metadata:
+            labels:
+              foo: bar
+```
+
+2. This exmaple restarts the deployment `staging/example-A` by updating the timestamp annotation on secret update.
+```yaml=
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: "..."
+spec:
+  rules:
+    - name: "..."
+      match:
+        resources:
+          kinds:
+          - Secret
+          names:
+          - certificate-store
+          namespaces:
+          - staging
+      mutate:
+        targets:
+        - kinds:
+          - Deployment
+          names:
+          - example-A
+          namespaces:
+          - staging
+        patchStrategicMerge:
+          metadata:
+            annotations:
+              kyverno.io/since-last-update: "{{ time_since('', '{{ request.object.metadata.creationTimestamp }}', '') }}"
+```
+
+3. This exmaple adds label `foo=bar` to both incoming and existing deployments on policy updates.
+```yaml=
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: "..."
+spec:
+  rules:
+    - name: "..."
+      match:
+        resources:
+          kinds:
+          - Deployment
+      mutate:
+        # mutateExisting is used to mutate the same 
+        # resources as specified in the match block
+        # the trigger resource is clusterpolicy/policy only
+        mutateExisting: true
+        # targets:
+        patchStrategicMerge:
+          metadata:
+            annotations:
+              foo: bar
+```
+
+4. This exmaple updates the threshold in the configmap upon pod creation.
+
+```yaml=
+#WIP
+
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: "..."
+spec:
+  rules:
+    - name: "..."
+      match:
+        resources:
+          any:
+          - kinds:
+            - Pod
+            namespaces:
+            - staging-A
+            - staging-B
+          - kinds:
+            - Namespace
+            namespaces:
+            - staging-*
+      preconditions:
+        any:
+        - key: "{{request.operation}}"
+          operator: Equals
+          value: CREATE
+      context:
+      - name: podCount
+        apiCall:
+          #urlPath: "/api/v1/namespaces/{{request.namespace}}/resourcequotas"
+          #jmesPath: "items | length(@)"   
+      mutate:
+        targets:
+        - kinds:
+          - ConfigMap
+          names:
+          - quota-threshold
+          namespaces:
+          - {{request.namespace}}
+        patchStrategicMerge:
+          data:
+            # threshold:
+```
+
+
+## Implementation
+
+### Watch Trigger Resources
+
+There are two ways to "watch" the trigger resource's activities:
+
+#### 1. Admission Webhook
+
+This approach leverages existing logic to register trigger resources in the webhook to receive events. It requires minimum code change and is consistent with other components, i.e., the Generate Controller.
+
+#### 2. Event Watcher
+
+An alternative is to register event watchers for trigger resources dynamically and de-register when a policy is removed. The solution lacks information about user context contained in the admission requests.
+
+Based on the above analysis, we will use the admission webhook to watch trigger resources.
+
+### Apply Policies
+
+Kyverno will mutate target(existing) resources in the background, not through the webhook. Since the trigger resources are watched via admission webhook (all instances will be active in HA), we need to create an itermediate resource to store the change request, and reconcile it in the background to apply such policies, similarly to how the generate controller reconciles generate requests. Once an event is received in the admission webhook, Kyverno will block the request until it creates the change request to the cluster. In this way, Kyverno is able to recover from failures in case it crashes in between.
+
+### Mutate Existing Resources
+
+The dynamic client will be used to fetch and upadate the target resources.
--- a/README.md
+++ b/README.md
@@ -9,3 +9,6 @@ New proposals should be submitted as a PR:
    https://github.com/kyverno/KDP/pulls
 
 
+# Tips
+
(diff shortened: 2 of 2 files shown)


proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+        # the trigger resource is clusterpolicy/policy only
+        mutateExisting: true
+        # targets:
+        patchStrategicMerge:
+          metadata:
+            annotations:
+              foo: bar
+```
+
+4. This exmaple updates the threshold in the configmap upon pod creation.
```
realshuting, line 143: @JimBugwadia - can you help further clarify this use case? Or is there any link for this request?

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+    + [Mutate Existing Resources](#mutate-existing-resources)
+
+<small><i><a href='http://ecotrust-canada.github.io/markdown-toc/'>Table of contents generated with markdown-toc</a></i></small>
+
+
+## Overview
+
+Kyverno webhook controller runs as an admission controller in Kubernetes, and it mutates resources upon admission review. Such mutation is performed based on resources operations (CREATE/UPDATE/DELETE) via webhook.
+
+As of Release 1.6.1, Kyverno has growning supprot for mutating existing resources:
```
chipzoller, line 25: "has growing support"

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+- The patches will be added as an annotation to existing resources.
+
+
+
+
+## Examples
+
+1. This exmaple adds the label `foo=bar` to deployment `staging/example-A` on configmap `staging/dictionary` update.
+
+```yaml=
```
chipzoller, line 50: When you use `=` outside the fence, it prevents syntax highlighting.

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+          - example-A
+          namespaces:
+          - staging
+        patchStrategicMerge:
+          metadata:
+            labels:
+              foo: bar
+```
+
+2. This exmaple restarts the deployment `staging/example-A` by updating the timestamp annotation on secret update.
```
chipzoller, line 87: "example"

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+          - example-A
+          namespaces:
+          - staging
+        patchStrategicMerge:
+          metadata:
+            annotations:
+              kyverno.io/since-last-update: "{{ time_since('', '{{ request.object.metadata.creationTimestamp }}', '') }}"
+```
+
+3. This exmaple adds label `foo=bar` to both incoming and existing deployments on policy updates.
```
chipzoller, line 118: "example"

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,213 @@
...
+
+Based on the above analysis, we will use the admission webhook to watch trigger resources.
+
+### Apply Policies
+
+Kyverno will mutate target(existing) resources in the background, not through the webhook. Since the trigger resources are watched via admission webhook (all instances will be active in HA), we need to create an itermediate resource to store the change request, and reconcile it in the background to apply such policies, similarly to how the generate controller reconciles generate requests. Once an event is received in the admission webhook, Kyverno will block the request until it creates the change request to the cluster. In this way, Kyverno is able to recover from failures in case it crashes in between.
+
+### Mutate Existing Resources
+
+The dynamic client will be used to fetch and upadate the target resources.
```
chipzoller, line 213: "update"

chipzoller:
What I don't see here is a pseudo policy on the background mutation use case (i.e., where there is no triggering resource). I think that's the primary use case for background mutation here. Couple suggestions on that:

1. needs to be a way in a policy to differentiate select between "new" resources which can be mutated (`CREATE` or `UPDATE`) and "old" resources which should be mutated (those which already exist)
2. needs parallel support in the Kyverno CLI with the `apply` command, similar to background generation, to allow users to understand the impact such a policy may have on existing resources.

realshuting:
@chipzoller - sorry I missed your comments. For some reason the notification goes into spam :(

To answer your questions:

>1. needs to be a way in a policy to differentiate select between "new" resources which can be mutated (CREATE or UPDATE) and "old" resources which should be mutated (those which already exist)

1.1. ""new" resources which can be mutated" - can be defined in `spec.rules.mutate.targets`;

1.2. ""old" resources which should be mutated (those which already exist)" - I assume old and new resources are the same in this case? If so, such use cases can be addressed by example 3, when the policy is installed, Kyverno will mutate existing resources.

>2. needs parallel support in the Kyverno CLI with the apply command, similar to background generation, to allow users to understand the impact such a policy may have on existing resources.

Good point! I've added it in the requirement.

chipzoller:
> @chipzoller - sorry I missed your comments. For some reason the notification goes into spam :(
> 
> To answer your questions:
> 
> > 1. needs to be a way in a policy to differentiate select between "new" resources which can be mutated (CREATE or UPDATE) and "old" resources which should be mutated (those which already exist)
> 
> 1.1. ""new" resources which can be mutated" - can be defined in `spec.rules.mutate.targets`;

But doesn't `mutate.targets` imply the use case where the target is different from the trigger? What if that's not the case and trigger = target? Seems that the existing way to write a mutate policy would suffice here for "new" resources.

> 1.2. ""old" resources which should be mutated (those which already exist)" - I assume old and new resources are the same in this case? If so, such use cases can be addressed by example 3, when the policy is installed, Kyverno will mutate existing resources.

I think there needs to be a clear way to distinguish between the two. There are three permutations here:

1. I want to mutate only "new" (incoming) resources but not pre-existing ones.
2. I want to mutate both "new" and "old" resources.
3. I want to mutate only "old" resources.

realshuting:
>1. I want to mutate only "new" (incoming) resources but not pre-existing ones.

The existing mutate policy can support this.

>2. I want to mutate both "new" and "old" resources.

2.1 "new" and "old" are the same - example 3, setting `mutateExisting=true`

2.2 "new" and "old" are different - example 1 or 2

>3. I want to mutate only "old" resources.

Good point, we can let the user set `mutateExisting=false` to support this. And we set the default value to nil/null.

chipzoller:
I think we also should generate an event on existing resources if they were mutated.

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+- Kyverno will fetch and mutate the existing resources once the trigger resource changes, the user needs to grant Kyverno permissions to operate on target resources.
+- The patches will be added as an annotation to existing resources.
+- CLI support with the apply command, similar to background generation, to allow users to understand the impact such a policy may have on existing resources.
+
+
+
+
+## Examples
+
+1. This example adds the label `foo=bar` to deployment `staging/example-A` on configmap `staging/dictionary` update.
```
chipzoller, line 49: Wouldn't this require a precondition to only match on `UPDATE`?

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+        # if this tag is specificied, Kyverno will 
+        # automatically perform post mutation, there's 
+        # no need to enable this feature additionally
+        targets:
+          # starts with GVK, can be extended to suppport
+          # labels, annotations, etc
+        - kinds:
+          - Deployment
+          names:
+          - example-A
```
chipzoller, line 79: We'd probably need to support variables from JMESPath expressions in `targets.names` and `targets.namespaces` taken from context lookups.

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+
+```yaml
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: "..."
+spec:
+  rules:
+    - name: "..."
+      match:
```
chipzoller, line 59: Since we're deprecating non `any`|`all` blocks under `match` and `exclude` may want to add this to ensure it's tested. Since we have to hang onto them for 1.7.0, it should work either way. Minor point but we've seen things break recently.

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+        # the trigger resource is clusterpolicy/policy only
+        mutateExisting: true
+        # targets:
+        patchStrategicMerge:
+          metadata:
+            annotations:
+              foo: bar
+```
+
+Setting `mutateExisting=false` will mutate existing resources but not incoming resources. The default value of `mutateExisting` is set to null.
```
chipzoller, line 144: Wouldn't `mutateExisting: false` just be the same as `mutateExisting: null` in that existing resources WOULD NOT be mutated? So maybe the default should just be `false`.

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+
+*The **trigger** resource is matched in the policy while the **target** resource is the resource to be mutated*.
+
+
+## Proposal
+
+To support mutate existing resources:
+
+- `spec.rules.match` will be used to identify operations on the trigger resources
+    - To *"mutate trigger resources based on policy update"*, the syntax is slightly different in order to simplify policy definition, see example 3 below.
```
chipzoller, line 38: This line is indented under the one above which says `spec.rules.match`, but in Example 3 I don't see a different syntax for that object, only the `mutate.mutateExisting` key. Are you envisioning the contents of `match` would need to be different?

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+- A new attribute `spec.rules.mutate.targets` will be introduced to specify a list of target resources to be mutated.
```
chipzoller, line 39: So is the idea when the `targets` object is present, anything under `mutate` will apply only to the resources selected by the filter under `targets` but NOT to the incoming (triggering) resource? If answer is "yes", do we want to allow a scenario in which, in a single rule, the body under `mutate` would apply to BOTH the targets AND the triggering resources?

proposals/mutate_existing_resources.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+## Overview
+
+Kyverno webhook controller runs as an admission controller in Kubernetes, and it mutates resources upon admission review. Such mutation is performed based on resources operations (CREATE/UPDATE/DELETE) via webhook.
+
+As of release 1.6.1, these are still the outstanding issues which Kyverno needs to support hence this proposal:
+
+- Mutate target resource which is different from the trigger resource, [#2139](https://github.com/kyverno/kyverno/issues/2139), [#1722](https://github.com/kyverno/kyverno/issues/1722).
+- Mutate trigger resources based on policy update, [#1607](https://github.com/kyverno/kyverno/issues/1607).
```
chipzoller, line 28: May want to add https://github.com/kyverno/kyverno/issues/1722 since you call this out in your first message.

proposals/mutate_existing_resources.md
realshuting: >So is the idea when the targets object is present, anything under mutate will apply only to the resources selected by the filter under targets but NOT to the incoming (triggering) resource?

Yes, see the revised version of the first point.

>do we want to allow a scenario in which, in a single rule, the body under mutate would apply to BOTH the targets AND the triggering resources?

I've thought about this, but it could easily result in an endless mutate loop. Even with the current proposal, we need to add policy validations carefully to prevent such an issue.

proposals/mutate_existing_resources.md
realshuting: Good catch, the original message is a bit confusing. Clarified and updated.

proposals/mutate_existing_resources.md
realshuting: Yes, rephrased it.

proposals/mutate_existing_resources.md
realshuting: Updated!

proposals/mutate_existing_resources.md
realshuting: Added!

proposals/mutate_existing_resources.md
realshuting: Yes, JMESPath can be used with the current implementation.

proposals/mutate_existing_resources.md
realshuting: Sorry I didn't get this, where do we want to add 1722?

realshuting:
>Should we provide for distinction between use cases 1 and 2? Use case for 1 is a one-time mutation whereas use case for 2 is a continual mutation, similar to a generate rule with sync: true enabled.

Just a clarification of the current proposal, there's no support as of now to mutate resources once. Like the existing mutate policies, resources will be mutated whenever there's an admission operation. The only difference with this feature is that the trigger can be from two sources:

- trigger resource events via admission requests
- policy events

If one-time mutation is something we want to support, we can extend this flag `spec.rules.mutate.mutateExisting` to accept different modes, say "Always" and "Once",  instead of `true` and `flase`.

realshuting:
> Also at least one use case that we might want to call as out of scope:
>
> Rolling back mutations in any of the above use cases

This is an interesting use case, I'm not sure if the previous state is stored and we can use that for rolling back since mutation happens during admission review. Do you know if `kubectl rollout` can revert admission review changes?

> Generate an event on existing resources if they were mutated.

Yes, will support.

>Have a new (multiple?) metric corresponding to existing resources that were mutated if not covered by a current metric.

Good catch, will need to check this during implementation.

proposals/mutate_existing_resources.md
realshuting: Yes, it's added here:
>- Mutate target resource which is different from the trigger resource, [#2139](https://github.com/kyverno/kyverno/issues/2139), [#1722](https://github.com/kyverno/kyverno/issues/1722).

Is this what you meant?

chipzoller: Yes I think so.

realshuting:
I've updated `mutate.target` structure to re-use the same target in generate policies. Otherwise changes are completed.

Can we merge this @prateekpandey14 @chipzoller @JimBugwadia ?

chipzoller:
I think so. I have resolved my comments.

realshuting:
@prateekpandey14 @JimBugwadia - needs another approval :)

JimBugwadia:
> > Should we provide for distinction between use cases 1 and 2? Use case for 1 is a one-time mutation whereas use case for 2 is a continual mutation, similar to a generate rule with sync: true enabled.
> 
> Just a clarification of the current proposal, there's no support as of now to mutate resources once. Like the existing mutate policies, resources will be mutated whenever there's an admission operation. The only difference with this feature is that the trigger can be from two sources:
> 
> * trigger resource events via admission requests
> * policy events
> 
> If one-time mutation is something we want to support, we can extend this flag `spec.rules.mutate.mutateExisting` to accept different modes, say "Always" and "Once", instead of `true` and `flase`.

I don't follow....if the user wants to apply the rule on existing resources in the cluster, will we not apply it as a background operation when the rule is created or modified?

chipzoller:
@realshuting what do we do with this now it's implemented yet the PR was never merged? Do we merge for posterity's sake or close?

realshuting:
> @realshuting what do we do with this now it's implemented yet the PR was never merged? Do we merge for posterity's sake or close?

I'll update to the current implementation today and we can then merge.

PR #18, opened 2022-05-03
Add requestTypes tag support

Prateeknandle:
Signed-off-by: Prateeknandle <prateeknandle@gmail.com>

### DIFF
--- a/proposals/add_requestTypes_tag_support
+++ b/proposals/add_requestTypes_tag_support
@@ -14,7 +14,7 @@ PR: [#3619](https://github.com/kyverno/kyverno/pull/3619)
 
 ## Introduction
 
-Until now we use `request.operation` under `condition.key`, to filter a request, after this proposal we'll use a tag `requestTypes` in `match` or `exclude` block to match/exclude specific requests coming from admission review requests.
+Until now we use `request.operation` under `preconditions[]`, to filter a request, after this proposal we'll use a tag `requestTypes` in `match` or `exclude` block to match/exclude specific requests coming from admission review requests.
 
 ## Proposal
 
--- a/proposals/add_requestTypes_tag_support
+++ b/proposals/add_requestTypes_tag_support
@@ -1,0 +1,54 @@
+# Kyverno Design Proposal - Adding Request Types to match and exclude
+
+Author: [Prateek Nandle](https://github.com/Prateeknandle)
+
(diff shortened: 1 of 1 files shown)


proposals/add_requestTypes_tag_support
```diff
@@ -0,0 +1,54 @@
...
+
+- [Kyverno Design Proposal - Adding Request Types to match and exclude](#kyverno-design-proposal---adding-request-types-to-match-and-exclude)
+  * [Introduction](#introduction)
+  * [Proposal](#proposal)
+  * [Example](#example)
+  * [Implementation](#implementation)
+
+## Introduction
+
+Until now we use `request.operation` under `condition.key`, to filter a request, after this proposal we'll use a tag `requestTypes` in `match` or `exclude` block to match/exclude specific requests coming from admission review requests.
```
chipzoller, line 17: This would be `preconditions[]` as it could be in key or value.

chipzoller (commented):
What will the default behavior be if `requestTypes` is not present? This change must allow backwards compatibility with existing policies which do not specify `requestTypes` and use preconditions.

Prateeknandle:
> What will the default behavior be if `requestTypes` is not present?

If `requestTypes` is not defined, Policy will behave same as now behaves when we don't filter out the request using `request.operation`, @realshuting is it correct statement?

> This change must allow backwards compatibility with existing policies which do not specify `requestTypes` and use preconditions.

`requestTypes` is a separate field, it'll not come under `any/all`, [code](https://github.com/kyverno/kyverno/pull/3619/files#diff-4161872653e054be4b55c2477d73cce4f0333b5123d0840b7f13ea6d5879df50R17).

proposals/add_requestTypes_tag_support
Prateeknandle: updated

chipzoller:
See https://github.com/kyverno/kyverno/issues/3110 for an example of where even this behavior internally isn't consistent depending on the contents of the rule. `requestTypes` should improve upon such behavior.

> `requestTypes` is a separate field, it'll not come under `any`/`all`, code.

Yes, that's fine, but my point was this addition must preserve operation of current policies if it is absent.

chipzoller:
This has been implemented in https://github.com/kyverno/kyverno/pull/6658

PR #25, opened 2022-07-30
Clean-up controller

chipzoller:
Proposes an idea for a clean-up controller for Kyverno.


chipzoller:
Implementation detail:  It may be better to use labels rather than annotations so we can configure a rule with an object selector for specifically those labels while not tampering with other rules in the same webhooks.

JimBugwadia (approved):
LGTM!

proposals/cleanup.md
```diff
@@ -0,0 +1,246 @@
...
+
+
+# Proposal
+
+In this proposal, Kyverno may function as a clean-up controller allowing it to identify and scavenge resources. There are two proposed capabilities which would be used to effect a clean-up.
+
+1. Use of two specific annotations which may be set on any resource:
+   - `kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the annotated resource was either created or when the annotation was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `kyverno.io/ttl: 30m`, `kyverno.io/ttl: 30d`
+   - `kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `kyverno.io/expires: 2022-08-04T00:30:00Z`, `kyverno.io/expires: 2022-09-30` 
+1. Use of a new rule type tentatively called `cleanup` which will execute in background mode on a schedule given in cron format. The rule will reuse existing Kyverno policy expression paradigms and capabilities to minimize introduced complexity.
```
prateekpandey14, line 52: 2. ?

chipzoller:
@realshuting see latest updates in the Implementation section. I tried to capture ideas proposed up to this point.

proposals/cleanup.md
```diff
@@ -0,0 +1,256 @@
...
+1. Explicit, per-resource clean-up declaration which may be set by a user either upon creation or retroactively.
+
+2. Implicit, bulk clean-up of resources on a scheduled basis which meet a certain criteria.
+
+
+# Proposal
+
+In this proposal, Kyverno may function as a clean-up controller allowing it to identify and scavenge resources. There are two proposed capabilities which would be used to effect a clean-up.
+
+1. Use of two specific annotations which may be set on any resource:
```
realshuting, line 49: We will need labels to register the webhookconfigurations to receive admission requests for matching resources. Note that the label value does not accept `:`.

realshuting (commented):
Looks good! Had a few questions beblow:

proposals/cleanup.md
```diff
@@ -0,0 +1,246 @@
...
+    c. Must only work in background mode and therefore must ensure background=true for this rule type
+    d. Must use existing Kyverno match/exclude and other policy authoring mechanisms to define logic
+    e. Must support clean-up of both Namespaced and cluster-scoped resources (including custom)
+    f. Must only clean up resources which have neither ttl nor expires annotations assigned
+    g. Must show, in some mechanism OTHER THAN the Kyverno log, any and all resources cleaned up by this rule. This should be at least events written to the policy containing the rule responsible for the clean up action.
+    h. Must support testing in the Kyverno CLI
+    i. Should be capable of checking and reporting, as a warning back to the user at time of policy creation, if Kyverno does not possess the necessary permissions to clean up the resources as defined in the rule's match block.
+    j. Should be able to determine unmounted PVs and unreferenced PVCs
+    k. Should be able to determine Pods not behind a Service
```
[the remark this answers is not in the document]
chipzoller: That's the idea. The new rule would be responsible for this, not the annotations/labels.

proposals/cleanup.md
```diff
@@ -0,0 +1,246 @@
...
+    l. Should support a dry run functionality so users may determine what resources would be cleaned up at the time the rule executes (Kyverno apply command)
```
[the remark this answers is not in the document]
chipzoller: No, this would be handled by the Kyverno CLI.

proposals/cleanup.md
chipzoller: Yes, "labels" = "annotations" in this proposal. As discussed, we probably need to use labels for the `ttl` and `expires` functionalities so a separate rule in the ValidatingWebhookConfiguration can be created which only matches on those two specific labels.

chipzoller:
@realshuting I've resolved conversations. If all looks good, will you approve?

proposals/cleanup.md
```diff
@@ -0,0 +1,256 @@
...
+
+# Motivation
+
+The initial idea for this feature came from issue [3483](https://github.com/kyverno/kyverno/issues/3483) authored by user luisdavim.
+
+As long as there has been Kubernetes, there has been the inevitability that some resources--most especially those created directly by human users--are left derelict at some point without a good way to identify what and where they are. As Kubernetes continues to expand with its myriad of possible custom resources, this problem grows over time. The ultimate problems this poses are few: 1) Kubernetes' key-value store, which has fairly limited storage capacity, may become overwhelmed by resources which should be deleted; and 2) these resources, in the case of Pods, may result in consumption of excess resources driving up cost and reducing workload consolidation abilities. The solution to these problems is to identify the derelict resources and remove (i.e., delete) them.
+
+There are two main use cases associated with this.
+
+1. Explicit, per-resource clean-up declaration which may be set by a user either upon creation or retroactively.
```
MarcelMue, line 40: I am concerned about this option - maybe I can't find it but how is implementation planned here?

This sounds (to me) like kyverno would need to detect the addition of a label or annotation on _any_ possible resource type. If this is indeed the intention, then efficient implementation in large clusters could be very hard.
Adding a watch to all possible resources wouldn't be a good idea, but getting the annotation in the minimal timeframe described here (1 minute) also sounds infeasible otherwise.

Maybe I am misunderstand this part of the KDP though.

MarcelMue (commented):
I only have concerns about the per-resource cleanup. Rule driven cleanup sounds like a nice idea to me :+1:

proposals/cleanup.md
chipzoller: >maybe I can't find it but how is implementation planned here?

Implementation (non-functional requirement) is intentionally open ended here. There are several options. I'm only trying to capture what the outward behavior should be (functional requirement).

>This sounds (to me) like kyverno would need to detect the addition of a label or annotation on any possible resource type. If this is indeed the intention, then efficient implementation in large clusters could be very hard.
Adding a watch to all possible resources wouldn't be a good idea, but getting the annotation in the minimal timeframe described here (1 minute) also sounds infeasible otherwise.

The idea that we (with @realshuting) had here was to use two labels. For example, `kyverno.io/ttl` and `kyverno.io/expires`. With these labels, we can have basically a "static" rule in the ValidatingWebhookConfiguration that uses the `objectSelector` on these labels causing the API server to only forward resources with one of these labels. Once Kyverno can observe this as an AdmissionReview, it can take some action (which is where the implementation details come into play).

realshuting (approved):

proposals/cleanup.md
MarcelMue: I hadn't considered the `objectSelector` usage - my only concern left is that this implies that Kyverno would register additional webhooks in the kubernetes API and therefor possibly get events multiple times from different registered webhooks. Not sure how big that problem really is though - maybe something to be discovered during testing.

chipzoller: It technically wouldn't even need to be a separate ValidatingWebhookConfiguration; it could just be a separate entry in the existing one. I think we'll have to work through these details when it gets implemented.

PR #26, opened 2022-08-01
update design of yaml signing and verification

rurikudo:
Signed-off-by: Riko Kudo <rurikudo@ibm.com>

- update the design of `validate.manifests` rule to align with current [code](https://github.com/rurikudo/kyverno/tree/dev/yaml-signing-sigstore)
- add the description for testing `validate.manifests` with CLI

### DIFF
--- a/proposals/yaml_signing_and_verification.md
+++ b/proposals/yaml_signing_and_verification.md
@@ -187,7 +187,6 @@ Kyverno Policy CRDs (Kyverno `Policy` and `ClusterPolicy`) is extended to includ
       - `annotations`
       - `repository`
   - `ignoreFields` 
-  - `skipUsers`
   - `repository`
   - `annotationDomain`
   - `dryrun`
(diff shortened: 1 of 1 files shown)


proposals/yaml_signing_and_verification.md
```diff
@@ -174,13 +177,18 @@ The priority would be to investigate the 2nd option.Resources to test with
...
+    - `count`  
+      `entries`
+      - `keys`
+      - `certificates`
+      - `keyless`
+      - `attestor`
+      - `annotations`
+      - `repository`
   - `ignoreFields` 
   - `skipUsers`
```
JimBugwadia, line 190: Since a Kyverno policy has the ability to exclude users in `rules.exclude.subjects`, do we need `skipUsers`:


```
❯ kubectl explain cpol.spec.rules.exclude.subjects
KIND:     ClusterPolicy
VERSION:  kyverno.io/v1

RESOURCE: subjects <[]Object>

DESCRIPTION:
     Subjects is the list of subject names like users, user groups, and service
     accounts.

     Subject contains a reference to the object or user identities a role
     binding applies to. This can either hold a direct API object reference, or
     a value for non-objects such as user and group names.

FIELDS:
   apiGroup     <string>
     APIGroup holds the API group of the referenced subject. Defaults to "" for
     ServiceAccount subjects. Defaults to "rbac.authorization.k8s.io" for User
     and Group subjects.

   kind <string> -required-
     Kind of object being referenced. Values defined by this API group are
     "User", "Group", and "ServiceAccount". If the Authorizer does not
     recognized the kind value, the Authorizer should report an error.

   name <string> -required-
     Name of the object being referenced.

   namespace    <string>
     Namespace of the referenced object. If the object kind is non-namespace,
     such as "User" or "Group", and this value is not empty the Authorizer
     should report an error.
```

rurikudo: @JimBugwadia  I confirmed that we don't need skipUsers field. We can reuse `spec.rules.exclude.subjects`.
I fixed the KDP and [PR Yaml signing and verification #4235](https://github.com/kyverno/kyverno/pull/4235).
Thank you.

samj1912 (approved):

PR #30, opened 2022-08-04
Remove deprecated CRD types and apiextensions.JSON

vyankyGH:
Proposes an idea for Remove deprecated CRD types and apiextensions.JSON

### DIFF
--- a/proposals/Remove deprecated CRD types and apiextensions.JSON.md
+++ b/proposals/Remove deprecated CRD types and apiextensions.JSON.md
@@ -1,0 +1,74 @@
+# Kyverno Design Proposal - Remove deprecated CRD types and apiextensions.JSON
+
+- Name: Remove deprecated CRD types and apiextensions.JSON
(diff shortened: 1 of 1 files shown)


chipzoller (approved):
Let's make sure that as part of this we're printing deprecation warnings for the fields removed in v2 schema if a user submits a v1 policy which uses these deprecated/removed fields. Also need to figure out how Kyverno upgrades would be handled if users have v1 policies already in the cluster and they upgrade Kyverno to a version in which they cannot be upgraded/migrated internally.

realshuting:
> Let's make sure that as part of this we're printing deprecation warnings for the fields removed in v2 schema if a user submits a v1 policy which uses these deprecated/removed fields. Also need to figure out how Kyverno upgrades would be handled if users have v1 policies already in the cluster and they upgrade Kyverno to a version in which they cannot be upgraded/migrated internally.

+1

realshuting (approved):

PR #33, opened 2022-09-16
policy exceptions

Eileen-Yu:
Signed-off-by: Eileen <eileenylj@gmail.com>

This KDP aims to address:
- https://github.com/kyverno/kyverno/issues/2627

Preview:
- https://github.com/Eileen-Yu/KDP/blob/397372abc8/proposals/allow_policy_exception.md

### DIFF
--- a/proposals/allow_policy_exception.md
+++ b/proposals/allow_policy_exception.md
@@ -1,0 +1,247 @@
+# Allowing Policy Exception
+
+- **Authors**: Jim Bugwadia (jim@nirmata.com), Eileen Yu (@Eileen-Yu)
+- **Date**: Sep 16th, 2022
+- **Update**:
+- **Abstract**: Overwrite policy for certain resources
+
+## Contents
+
+- [Contents](#contents)
+- [Introduction](#introduction)
+- [The Problem](#the-problem)
+- [Proposed Solution](#proposed-solution)
+  - [Possible Solution](#possible-solution)
+    - [Request](#request)
+    - [Response](#response)
+  - [Possible User Stories](#possible-user-stories)
+- [Next Steps](#next-steps)
+
+## Introduction
+
+This document introduces a new CRD `PolicyViolationExceptions` . The proposed CRD aims to provide certain resource with privileges. The resource approved by such CRD would temporarily escape from specific policy rules.
+
+## The Problem
+
+When a policy is applied, all the matching resources / users have to follow the rules specified in it. But in certain cases, user may find necessary to get permission which will break current rules. Nonetheless, the original rule is still valid and should not be changed, especially for those cases that would break the rule for only once or in a period. This requires to introduce the exception mechanism.
+
+Some related issues on this:
+
+- [Overwriting Existing policies](https://github.com/kyverno/kyverno/discussions/4310)
+- [Feature: New CRDs for PolicyViolationExcepotions](https://github.com/kyverno/kyverno/issues/2627)
+
+One possible solution is to overwrite the existing policies, where the user adds certain cases in the `exclude` spec, or retune some other fields. However, this would cause some ambiguous permission boundary as well as some security risks. Also, additional spec would be needed to specify if a certain policy can be overwrote or not.
+
+It is recommended to introduce a new CRD for those privileged resources while avoid affecting the existing policy.
+
+## Proposed Solution
+
+### Possible Solution
+
+For a clearer authority management, a `Request-Response` mechanism can be considered.
+
+#### Request
+
+The user would create the `PolicyViolationExceptions` CR.
+
+`PolicyViolationExceptions` Example:
+
+```yaml
+...
+spec:
+  policyName:..
+  ruleName:..
+  reason:..
+  exclude:..
+  duration(optional):..
+status:
+  state: `PendingApproval` / `Approved` / `Rejected`
+```
+
+- policyName: target policy name
+- ruleName: target rule name
+- reason: brief explanation for the exception
+- exclude: specify the resource that need to be excluded from the target policy, may include resource information, e.g. kind, name, namespace, labels
+- duration(optional): how long the exception would be valid
+
+#### Response
+
+Once the CR is created, the admin would be able to review the content and modify it if necessary.
+
+1. If the exception is approved, a new CR `Approval` would be generated.
+
+`Approval` Example:
+
+```yaml
+---
+spec:
+  exceptions: name:..
+    nameSpace:..
+    duration:..
+```
+
+- name: name for the approved exception
+- nameSpace: valid namespace that would be affected by the exception
+- duration: how long the exception would be valid
+
+Then the relative resource would get mutated automatically based on the dual effect of the original policy and the exception.
+
+2. If the exception is rejected, then nothing would go on.
+
+The CRs would get recycled once the resources are reconciled OR it is outdated.
+
+### Possible User Stories
+
+**As a user, I want to be excepted from label tag rule for test.**
+
+**a. What policy blocks what resources/user?**
+
+The policy disables the user to use an image tag called `latest`.
+
+```yaml
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: disallow-latest-tag
+  annotations:
+    policies.kyverno.io/title: Disallow Latest Tag
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/severity: medium
+    policies.kyverno.io/subject: Pod
+    policies.kyverno.io/description: >-
+      The ':latest' tag is mutable and can lead to unexpected errors if the
+      image changes. A best practice is to use an immutable tag that maps to
+      a specific version of an application Pod. This policy validates that the image
+      specifies a tag and that it is not called `latest`.
+spec:
+  validationFailureAction: enforce
+  background: true
+  rules:
+    - name: validate-image-tag
+      match:
+        resources:
+          kinds:
+            - Deployment
+      validate:
+        message: "Using a mutable image tag e.g. 'latest' is not allowed."
+        pattern:
+          spec:
+            template:
+              spec:
+                containers:
+                  - image: "!*:latest"
+```
+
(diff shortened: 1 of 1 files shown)


chipzoller:
Hi @Eileen-Yu , please complete the DCO sign-off for this and all future commits as required by CNCF.

chipzoller (changes requested):

Eileen-Yu:
Since the KDP has been updated, mark all outdated comment as resolved.

proposals/allow_policy_exception.md
```diff
@@ -0,0 +1,358 @@
...
+  - name: name of corresponding `PolicyExceptionRequest`
+  - namespace: namespace of corresponding `PolicyExceptionRequest`
+  - status: whether admin decide to approve / deny the `PolicyExceptionRequest`
+  - reason: why admin approve / deny the `PolicyExceptionRequest`
+
+The `status` and `reason` would also be updated to the `PolicyExceptionRequest` to let the user know by the `PolicyExceptionController`.
+
+### Ideal Usage
+
+1. User creates a CR `PolicyExceptionRequest` to request for exception in a certain namespace.
```
chipzoller, line 114: The PER is a namespaced resource which makes sense, but two things we should consider:

1. How or if a user may request an exception for a cluster-scoped resource (ex., a Calico GlobalNetworkPolicy)
2. Protection so that a user in Namespace A cannot create an exception for a violating resource in Namespace B.

proposals/allow_policy_exception.md
```diff
@@ -0,0 +1,358 @@
...
+  - [Possible User Stories](#possible-user-stories)
+- [Implementation](#implementation)
+- [Alternative Solution](#alternative-solution)
+- [Next Steps](#next-steps)
+
+## Introduction
+
+Kyverno policy rules contain an `exclude` block which allows excluding admission requests by resource (name, API group, versions, and kinds), resource labels, namespaces, namespace labels, users, use groups, service accounts, user role and groups. This construct allows flexibility in selecting when a policy should be applied to a admission review request.
+
+This proposal introduces the concept of a `PolicyException` which is similar to the `exclude` declaration but is decoupled from the policy lifecycle.
```
chipzoller, line 24: Let's clearly state this proposal is only applicable to validate rule types. That is correct, yes?

proposals/allow_policy_exception.md
```diff
@@ -0,0 +1,358 @@
...
+  - namespace: namespace of corresponding `PolicyExceptionRequest`
+  - status: whether admin decide to approve / deny the `PolicyExceptionRequest`
+  - reason: why admin approve / deny the `PolicyExceptionRequest`
+
+The `status` and `reason` would also be updated to the `PolicyExceptionRequest` to let the user know by the `PolicyExceptionController`.
+
+### Ideal Usage
+
+1. User creates a CR `PolicyExceptionRequest` to request for exception in a certain namespace.
+2. Admin reviews the request, and adds new element in `PolicyExceptionApprovals` to approve / deny the request.
```
chipzoller, line 115: It isn't clear to me if you are proposing just one PEA per cluster or multiple. Can it be multiple?

proposals/allow_policy_exception.md
```diff
@@ -0,0 +1,358 @@
...
+3. A new `PolicyExceptionController` would update the `PolicyExceptionRequest` based on admin's feedback. If not approved, user can modify and resubmit `PolicyExceptionRequest`.
+4. The `PolicyExceptionApprovals` would be cached by Kyverno for policy match and exclude.
+5. User creates the corresponding resource. Kyverno matches the resource with corresponding policies and exceptions. If all broken rules can be matched with exceptions, let the request pass.
```
chipzoller, line 118: I would like to see somewhere in here the logs/events that will be written and where whenever a resource is allowed based upon an exception. This will be a critical piece for security teams and auditors.

chipzoller (commented):
Looks pretty good.

Is it also possible to record for the admin to see which user/ServiceAccount/group created the exception request?

proposals/allow_policy_exception.md
Eileen-Yu: Sounds good 👍 We update the KDP, now the PER can be either namespaced / cluster-wide.

proposals/allow_policy_exception.md
Eileen-Yu: We suppose it's not only limited to `validation`. 
These exceptions would be added to the policy rule `exclude` block in Kyverno finally, so should share similar usage with current `exclude`.

proposals/allow_policy_exception.md
Eileen-Yu: Based on discussion in weekly, we now would have multiple PEAs (now changed to `PolicyExceptionResponse`). 
Admin would create one and only one `PolicyExceptionResponse` for each `PolicyExceptionRequest`.

proposals/allow_policy_exception.md
Eileen-Yu: Sure, make sense! Have updated this part in KDP.

JimBugwadia:
@eddycharly we had started with a scheme like that but it seemed to lead to synchronous workflows.

It would be nice to simply have a PolicyException object and defer the approval lifecycle to external tools. 

However this requires that the PolicyException resource be secured and handled with care. Which implies that developers cannot create asynchronously, as part of their workloads.

One idea was to require signed exceptions as a protection mechanism - but that creates coupling to the YAML signing capabilities.

Perhaps we can capture this as an alternate design and discuss further.

eddycharly:
> However this requires that the PolicyException resource be secured and handled with care. Which implies that developers cannot create asynchronously, as part of their workloads.

Kyverno should only read `PolicyException`s and we could argue that the only one service account that should be able to create them is the controller responsible for approving `PolicyExceptionRequest`s.

I mean, not every organisation is going to need to this request/approval mechanism, this makes sense to me that it is not coupled with core kyverno.

Most of the time developers have limited permissions on a cluster, and if Argocd is used for example that's very easy to enforce permissions at argocd level too. I find `PolicyException`s useful but if I had to use it in a GitOps workflow I feel like it could be a pain. I would rather expect that `PolicyException`s are created from a specific repository where developers can open PRs and let Argo sync them automatically, GitOps means we don't need to connect to the cluster at all.

With this design it requires an Admin to create the approval when a request is made which is not what all users will want/need. It makes sense for some not for all, we should design that can adapt to different use cases while not restricting others ones IMHO.

JimBugwadia:
@eddycharly  - in this alternate approach, are you proposing 2 CRs for Kyverno (a request and a response) or one (an extremely approved exception)?

eddycharly:
@JimBugwadia from the kyverno POV only one `PolicyException` is enough.

The request/approval mechanism is separate and eventually, when a request is approved this will create the corresponding `PolicyException`. But for kyverno itself only `PolicyException`s matter.

stone-z:
Following up from [a Slack thread](https://kubernetes.slack.com/archives/CLGR9BJU9/p1667334617072309), just a brief comment about use cases for this feature. This is intentionally short, happy to expand or join a call to elaborate if needed.

I think this concept is great, and the right direction to be able to accommodate complex automation alongside Kyverno. There are some use cases, however, which I anticipate will be desired but not met by this design, or at least I haven't worked out exactly what the intended workflow would be. Things like:
- Use case: Admin wants to exclude a list of trusted images from PSS policies
  - Problem 1: `exclude` blocks can't reference the image used in a resource (only name/namespace and selectors)
  - Problem 2: `exclude` blocks can't use contexts, so no possibility to reference e.g. a list in a ConfigMap
  - Workaround: Admin creates individual exclusions (or static list) for Deployments instead of images
- Use case: Admin wants to create individual exclusions for Deployments from PSS policies ^
  - Problem 1: PSS policies are written at the Pod level, so:
    - Admin can't exclude them based on Deployment name/selectors
    - If excluding by Pod name (using a wildcard, since Pod names usually aren't known at admission time), autogen is disabled for that policy
  - Workaround: Admin must exclude Pod by name using a wildcard, leading to:
  - Problem 2: Users who can create Pods can now bypass PSS policies by creating a specially-named Pod

Suggestions:
- Support exclusions based on parent resource names/selectors (e.g. exclude Pod from PSS policy based on Deployment name)
- Support or somehow enable contexts+preconditions set via PERs

I might have missed something as I'm very much still trying to wrap my head around exceptions in a post-PSP, out-of-tree world. Happy to chat and expand more if needed. Thanks for such a great project!

JimBugwadia:
Hi @stone-z - thanks for the review and feedback!  Some thoughts and questions below:

> Use case: Admin wants to exclude a list of trusted images from PSS policies

Are you using the new `validate.podSecurity` rule or the individual pod security policies?

When excluding images from PSS controls, it seems like its best to be fine grained and exclude an image for a specific control. The `validate.podSecurity` rule allows that. 

What would be the benefit of managing the image & control pairs externally vs directly in the policy? 

>  Use case: Admin wants to create individual exclusions for Deployments from PSS policies

This should be possible. @Eileen-Yu - can you please share the sample for excluding a Deployment and the pods it controls?

JimBugwadia (approved):

JimBugwadia (approved):

Eileen-Yu:
Hi @stone-z  Thx for your feedback! 

For the second use case you mentioned, 
> * Use case: Admin wants to create individual exclusions for Deployments from PSS policies ^

Considering on the case of autogen, we would introduce some best practices for the users to specify exceptions.
As for the exclusions for Deployments, you can:
```yaml
---
apiVersion: kyverno.io/v2alpha1
kind: PolicyException
metadata:
  name: bypass-test-deployment
  namespace: kyverno
spec:
  exceptions:
    - policyName: require-labels
      ruleNames:
        - check-for-labels
        - autogen-check-for-labels
  match:
    any:
      - resources:
          kinds:
            - Deployment
          names:
            - nginx
          namespaces:
            - default
      - resources:
          kinds:
            - Pod
          namespaces:
            - default
          selector:
            matchLabels:
              app: nginx
```

chipzoller (approved):

Issue #14, opened 2022-04-15
[Feature] Allow calling in cluster services


samj1912:
### Problem Statement

There is currently no easy way to extend kyverno capabilities. Allowing kyverno to make calls to in-cluster services could unlock such capabilities.

### Solution Description

Allow a new `serviceCall` context variable type.

```yaml
rules:
- name: example-service-call
  context:
  # name of the variable to capture the service output from
  - name: output
    serviceCall:
      service:
        name: service-name
        namespace: some-namespace
        # optional
        port: 80
        # optional
        path: validate-namespace
        # optional url GET params
        params:
          namespace: '{{ request.object.metadata.namespace }}'
        # optional JSON post data. If defined.
        data:
          method: '{{ request.method }}'
        # http verb for the call. One of the http verb names. Defaults to GET if data is not defined otherwise POST
        verb: POST
      # optional timeout in seconds per retry, default is 2s
      timeout: 10
      # optional retries. Defaults to 1
      retries: 2
      # optional jmespath expression evaluated on the response before the variable is assigned 
      jmesPath: "items | length(@)"
```

The above service calls would limit the return accept type to json only. The encoding would also be json for the post data. 

We would require an in cluster `Service` object to be defined to reach out to.

### Research

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have searched other issues in this repository and mine is not recorded.

chipzoller:
Do you want to covert this to a KDP? Would you plan on developing this for 1.8.0 potentially?

Dentrax:
Dropping some use case about the proposal as answering of @JimBugwadia's comment: https://github.com/kyverno/kyverno/pull/3820#issuecomment-1191710517

> @Dentrax - are you looking for this to address a use case? If so, I would like to learn more about that.

Image you are running hundreds of clusters and want to create some _excluding_ policies using _ConfigMap_. For example, we want to exclude some critical images such as _istio_, _node-local-dns_ etc. But this task requires some stuff such creating ConfigMaps in every cluster, and we have to keep maintaining those.

I'd like to store my _context_ in a centralized config solution such as _consul_. By doing so, we are able to store all excluding rules in a single place for better maintainability and real-time operation for _upserting_ the new configs.

But I just noticed this issue is for _in-cluster services_ only. What I was looking for is that _external data_. (I actually mean _out-cluster_ arbitrary apiCall) Should we create another KDP for this?

```
rules:
[...]

PR #40, opened 2022-12-22
proposal for Service API calls 

JimBugwadia:
Signed-off-by: Jim Bugwadia <jim@nirmata.com>

Initial proposal at: https://github.com/kyverno/KDP/issues/14.

### DIFF
--- a/proposals/service_api_calls.md
+++ b/proposals/service_api_calls.md
@@ -117,7 +117,7 @@ FIELDS:
    data <[]Object>
      Data specifies the POST data sent to the server.
 
-   requestType  <string> -required-
+   method <string> -required-
      method is the HTTP request method (GET or POST).
 
    urlPath      <string> -required-
--- a/proposals/service_api_calls.md
+++ b/proposals/service_api_calls.md
@@ -1,0 +1,186 @@
+# Meta
+[meta]: #meta
+- Name: Service API Call
+- Start Date: December 20, 2022
+- Update data (optional):  December 20, 2022
+- Author(s): @JimBugwadia
+
+
+# Table of Contents
+[table-of-contents]: #table-of-contents
+- [Meta](#meta)
+- [Table of Contents](#table-of-contents)
+- [Overview](#overview)
+- [Definitions](#definitions)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+- [Implementation](#implementation)
+- [Migration (OPTIONAL)](#migration-optional)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Prior Art](#prior-art)
+- [Unresolved Questions](#unresolved-questions)
+- [CRD Changes (OPTIONAL)](#crd-changes-optional)
+
+# Overview
+[overview]: #overview
+
+Support HTTP/S GET and POST calls to JSON web services.
+
+# Definitions
+[definitions]: #definitions
+
+
+# Motivation
+[motivation]: #motivation
+
+Kyverno versions 1.9 and earlier support API calls to the Kubernetes API server and loading data from ConfigMap resources. 
+
+However, policies often require data from other services e.g. a metrics service. Currently, this requires writing custom code to periodically fetch data from the services and store the results in a ConfigMap or other Kubernetes resource that can be retrieved by Kyverno. This feature allows the service to be directly queried, eliminating the need to write custom services and create Kubernetes custom resources for policy data.
+
+In other cases, some policies may require delegation of processing to other controllers. For example, a separate controller may periodically fetch and cache external data which is used to process policies. With this feature, Kyverno can send the service a POST request with context data, and the service can respond back with results that are used to make the policy decision. 
+
+Finally, in some scenarios 3rd party library code or binaries need to be executed to make a policy decision. Currently, the only way to support these "extensions" is to directly update the Kyverno core. With this feature, it will be possible to write Kyverno extensions as separate services hence decoupling the Kyverno code from 3rd party libraries or binaries.
+
+# Proposal
+
+This feature enables:
+- HTTP GET and POST API calls to any service
+- Encryption and server authentication using a certificate bundle
+- Client authentication using Service Account token volume projection 
+- Passing context data to the external service
+- Applying JMESPath transformation to the results
+- Storing service API call results in the rule context for further processing
+
+# Implementation
+
+Here are the changes involved:
+1. Add a new `service` to the existing `APICall` declaration, to allow specifying information like the URL and CA bundle. The existing `
+2. Update the logic to handle APICall to check whether a `urlPath` is configured to invoke the API server. Otherwise, process the `service` declaration to execute an HTTP GET or POST to the JSON web service.
+3. Apply the JMESPath (if specified) to the results and store in context, as before. 
+
+
+## CRD Changes
+
+Changes for the APICall object:
+
+```sh
+ ❯ kubectl explain cpol.spec.rules.context.apiCall
+
+KIND:     ClusterPolicy
+VERSION:  kyverno.io/v1
+
+RESOURCE: apiCall <Object>
+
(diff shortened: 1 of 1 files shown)


proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+In other cases, some policies may require delegation of processing to other controllers. For example, a separate controller may periodically fetch and cache external data which is used to process policies. With this feature, Kyverno can send the service a POST request with context data, and the service can respond back with results that are used to make the policy decision. 
+
+Finally, in some scenarios 3rd party library code or binaries need to be executed to make a policy decision. Currently, the only way to support these "extensions" is to directly update the Kyverno core. With this feature, it will be possible to write Kyverno extensions as separate services hence decoupling the Kyverno code from 3rd party libraries or binaries.
+
+# Proposal
+
+This feature enables:
+- HTTP GET and POST API calls to any service
+- Encryption and server authentication using a certificate bundle
+- Client authentication using Service Account token volume projection
```
samj1912, line 50: @JimBugwadia I think we should also add client authentication using the `Authorization` header. Most of the web services and k8s itself use JWTs for authentication. Having it available as a way to authenticate would dramatically help with the usability of this feature.

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+
+KIND:     ClusterPolicy
+VERSION:  kyverno.io/v1
+
+RESOURCE: service <Object>
+
+DESCRIPTION:
+     Service is an API call to a JSON web service
+
+FIELDS:
```
samj1912, line 112: Can we also add a field called `Headers` here which allows setting the http headers?

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+    - name: failedCount
+      apiCall:
+        service:
+          url: https://service.namespace:443/validateImages
+          caBundle: {{ cert }}
+          requestType: POST
+          data:
+          - key: images
+            value: "{{ images.[containers, initContainers, ephemeralContainers][].*.reference[] }}"
+        jmesPath: "items[?status == 'failed'] | length(@)"
```
samj1912, line 158: I would suggest removing jmespath from the service field. We support variable reassignment and jmespath transformations using inline context variables. We can probably use that to do any transformations in a consistent fashion rather than introducing it here. I know we have this field in other external data calls, but all of them are half baked and miss features that inline variables offer like setting defaults in case values are missing.

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+
+FIELDS:
+   caBundle     <string>
+     CABundle is a PEM encoded CA bundle which will be used to validate the
+     server certificate.
+
+   data <[]Object>
+     Data specifies the POST data sent to the server.
+
+   requestType  <string> -required-
```
samj1912, line 120:
```suggestion
   method <string> -required-
```

proposals/service_api_calls.md
samj1912: In general I believe we should support the following methods of authentication -

- Basic
- JWT
- certs/mTLS

The above 3 are also supported by kubernetes.

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+
+DESCRIPTION:
+     Service is an API call to a JSON web service
+
+FIELDS:
+   caBundle     <string>
+     CABundle is a PEM encoded CA bundle which will be used to validate the
+     server certificate.
+
+   data <[]Object>
```
samj1912, line 117: Can we also add a field to specificy URL query parameters in addition to post/body? It will make the user experience easy and Kyverno can handle properly constructing/encoding the url params.

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Proposal
+
+This feature enables:
+- HTTP GET and POST API calls to any service
+- Encryption and server authentication using a certificate bundle
+- Client authentication using Service Account token volume projection 
+- Passing context data to the external service
+- Applying JMESPath transformation to the results
+- Storing service API call results in the rule context for further processing
```
samj1912, line 53: Can we also document the output that is stored? We will need to store the response headers, body, status code at least.

proposals/service_api_calls.md
samj1912: Example of how goreleaser implements this https://goreleaser.com/customization/upload/#username

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Implementation
+
+Here are the changes involved:
+1. Add a new `service` to the existing `APICall` declaration, to allow specifying information like the URL and CA bundle. The existing `
```
chipzoller, line 58: End of this line looks like it got cut off.

chipzoller (approved):
This looks good to me. I do like Sam's suggestions though.

samj1912:
We should add retries, timeouts and polling periods at the very least. See https://github.com/kyverno/KDP/issues/14

proposals/service_api_calls.md
```diff
@@ -0,0 +1,186 @@
...
+VERSION:  kyverno.io/v1
+
+RESOURCE: service <Object>
+
+DESCRIPTION:
+     Service is an API call to a JSON web service
+
+FIELDS:
+   caBundle     <string>
+     CABundle is a PEM encoded CA bundle which will be used to validate the
```
samj1912, line 114: Similar to OPA, kubernetes, we should introduce a insecureSkipSSLVerify flag.

proposals/service_api_calls.md
JimBugwadia: @samj1912 - yes, the intent was to add an API token to the HTTP Authorization header. I've clarified.

proposals/service_api_calls.md
JimBugwadia: Yes, its possible. I've noted in the future enhancements section.

proposals/service_api_calls.md
JimBugwadia: This is an existing attribute, under APICall. It is used for the Kubernetes API server calls, so was planning to also reuse for any other service call. Its optional.

proposals/service_api_calls.md
JimBugwadia: I've noted in the future enhancements section.

proposals/service_api_calls.md
JimBugwadia: Only the returned JSON data in the response body is stored. Added details to clarify.

proposals/service_api_calls.md
JimBugwadia: fixed, thanks!

proposals/service_api_calls.md
JimBugwadia: I've noted in the future enhancements section.

JimBugwadia:
Hi @samj1912 - thanks for the great suggestions! I agree with them, but would like to get a minimal working implementation available quickly, and then extend based on use cases.  The initial use case I am looking to address is invoking a service that runs the notation binary and plugins (the NotaryV2 CLI), and returns a response.

If you have other specific use cases in mind, that would be limited without initial support for some of the suggested options, please let me know.

samj1912:
@JimBugwadia thanks for the insight. Would we be willing to accept PRs for the features I suggested? We currently have a couple of validation webhooks that query some external services (that provide http wrappers over some data sources) to validate our custom resources. This results in additional custom webhooks and we haven't been able to port this logic to Kyverno because of a lack of this service call feature. I would be happy to contribute these features along with timeouts/retries.

JimBugwadia:
Hi @samj1912 - yes, absolutely! PRs to add these enhancements are very welcome, and I am happy to collaborate and help.

samj1912 (approved):

realshuting (approved):

PR #41, opened 2023-02-16
proposal for introducing ebpf

Eileen-Yu:
This proposal aims to enable users to write kyverno policies at a finer granularity by combining with ebpf.

There are some challenges when design, and many other parts need to cover. 
So currently I left them in `Open Question`, and would be glad if there is any suggestion.

Preview: 
- https://github.com/kyverno/KDP/blob/17b2d68ea08eb331b56543fbcc943db034491676/proposals/introduce_ebpf.md

### DIFF
--- a/proposals/introduce_ebpf.md
+++ b/proposals/introduce_ebpf.md
@@ -1,0 +1,74 @@
+# Introduce eBPF
+
+- **Authors**: Eileen Yu (@Eileen-Yu), Jim Bugwadia (jim@nirmata.com)
+- **Created**: Feb 15th, 2023
+- **Updated**: Feb 20th, 2023
+- **Abstract**: Introduce eBPF to trigger Kyverno policy for extended use cases
+
+## Contents
+
+- [Introduction](#introduction)
+- [The Problem](#the-problem)
+- [Proposed Solution](#proposed-solution)
+- [Open Question](#open-question)
+
+## Introduction
+
+Kyverno is now running as a dynamic admission controller for policy management. This allows users to apply policies to Kubernetes resources during admission controls and as background scans.
+
+This proposal aims to extend the ability of Kyverno, so that it can apply policies based on runtime events detected via eBPF events.
+
+## The Problem
+
+Besides configuration changes and periodic scans, security conscious users may want to trigger the application of policies based on runtime events . This goes far beyond simply "approve / deny" the request. One instance is, users may want to detect when a volume is mounted in a running container, via an operation that bypasses the API server, and shutdown a pod . Another example is to report a violation for certain network calls .
+
+It would be best if there is a way to enforce these security requirements as Kyverno policies.
+
+## Proposed Solution
+
+To instrument on the system call interface, some reliable eBPF observability tools (e.g Tetragon, Falco) can help to monitor which sys-call is invoked and generate logs / events / metrics.
+
+Once Kyverno gets the data through informer / websocket, it would check if a certain existing policy should be triggered. If so, Kyverno would apply rules to modify or delete the manifest of corresponding resource, and generate rule violation reports.
+
+## Open Question:
+
+**Q1: What type of logs / events / metrics that Kyverno would accept? How to parse them?**
+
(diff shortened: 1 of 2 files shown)


proposals/introduce_ebpf.md
```diff
@@ -0,0 +1,82 @@
...
+Kyverno is now running as a dynamic admission controller for policy management. This allows users to manage Kubernetes native resources.
+
+This proposal aims to extend the ability of kyverno by introduing ebpf, so that it could cover events at a finer granularity at runtime.
+
+## The Problem
+
+At current stage kyverno focus on enforcing policies based on admission webhook. This enable users to define policy using k8s manifests to easily manage resources.
+
+Well, it is common for users to raise further fine-grained requirements out of security concern. This goes far beyond simple "approve / deny" the request.
+One instance is, users may want to limit the volume mount to a specific container. Another example is to control the access to certain API paths based on layer 7 calls.
```
chipzoller, line 26: >users may want to limit the volume mount to a specific container

But this is totally possible today with a Kyverno validate rule, or is there some nuance here which isn't said?

chipzoller: >Another example is to control the access to certain API paths based on layer 7 calls

Can you explain what you mean here? "layer 7 calls" doesn't make sense to me in this context.

proposals/introduce_ebpf.md
```diff
@@ -0,0 +1,82 @@
...
+
+It would be better if there is a way to enforce these granular configurations as K8s policies without needing to learn new language.
+
+## Proposed Solution
```
chipzoller, line 30: I know this is (currently) marked as a WIP, but the proposed solution here I don't quite follow. It sounds like you're proposing something like

```mermaid
flowchart LR
    id1[eBPF Tool] --> id2[K8s Event] --> id3[Kyverno validate rule] --> id4[Action]
```

proposals/introduce_ebpf.md
```diff
@@ -0,0 +1,82 @@
...
+  ---
+  spec:
+    rules:
+      - name: restrict-volume-mount-to-container-A
+        match:
+          any:
+            - resources:
+              kinds:
+                - Pod
+        Restirct:
```
chipzoller, line 69: "Restrict"

proposals/introduce_ebpf.md
```diff
@@ -0,0 +1,82 @@
...
+At current stage kyverno focus on enforcing policies based on admission webhook. This enable users to define policy using k8s manifests to easily manage resources.
+
+Well, it is common for users to raise further fine-grained requirements out of security concern. This goes far beyond simple "approve / deny" the request.
+One instance is, users may want to limit the volume mount to a specific container. Another example is to control the access to certain API paths based on layer 7 calls.
+
+It would be better if there is a way to enforce these granular configurations as K8s policies without needing to learn new language.
+
+## Proposed Solution
+
+To instrument on the system call interface, some reliable ebpf obeservability tools (e.g Tetragon, Falco) can help to monitor which sys-call is invoked and generate logs / events / metrics.
```
chipzoller, line 32: "observability"

proposals/introduce_ebpf.md
```diff
@@ -0,0 +1,82 @@
...
+          message: Restrict volume mount to container A
+          pattern:
+            spec:
+              containers:
+                - name: A
+              behavior:
+                - volume-mount
+  ```
+
+  The `Restirct` spec means to block the target behavior.
```
chipzoller, line 79: "Restrict"

proposals/introduce_ebpf.md
Eileen-Yu: > But this is totally possible today with a Kyverno validate rule, or is there some nuance here which isn't said?

The main difference is we wish to block the `violation` at runtime, rather than check the static manifest to decide whether to apply it. So even if there are resources that `bypassed` the rule, they can still be caught and shut down.

Eileen-Yu: > Can you explain what you mean here? "layer 7 calls" doesn't make sense to me in this context.

Thx for pointing out! 

ebpf can be aware of network calls in the application level, compared with the network policy which mainly around TCP/IP. 

Updated in the docs.

JimBugwadia:
Closing this as there has been no activity.

PR #45, opened 2023-06-07
[Docs]: Fix the inconsistencies in the cleanup proposal

### LINKED ISSUE #44: [Docs]: Fix the inconsistencies in the cleanup proposal
As per the proposal overview, the cleanup policy's ability should be leveraged to the removal of the resources as per the labels, which are defined in typemeta, and we should implement the ability to remove the resources as per the labels defined. But, in the proposal it's mentioned to add annotation support instead of label.

## Expected Behaviour

We should remove the inconsistencies (replace annotations with labels, and similar things).

VedRatan:
Fixed the inconsistencies, and replaced the annotations with labels and similar fixes.

Closes #44

### DIFF
--- a/proposals/cleanup.md
+++ b/proposals/cleanup.md
@@ -252,9 +252,8 @@ Kyverno could alternatively leverage a CronJob resource to perform the deletions
 
 - Added the new validation-webhook to make sure that the resources with the label value follows the specific pattern, if not a warning is thrown to the user that the label to which the Kyverno is keeping a watch doesn't follow the recommneded format.
 
-- Created a webhook which will only keep a watch for the resources with the label `cleanup.kyverno.io/ttl` and verfies that the label value adheres to the format defined or not.
+- Created a webhook which will only keep a watch for the resources with the label `cleanup.kyverno.io/ttl` (used the labelSelector from the from the apimachinery/meta/v1 library to do so) and verfies that the label value adheres to the format defined or not.
 
-- Intr
 
 ## Pros And Cons of Both implementations
 **Implementation 1**: add a new cleanup controller in the Kyerno main process
--- a/proposals/cleanup.md
+++ b/proposals/cleanup.md
@@ -230,6 +230,10 @@ Kyverno could alternatively leverage a CronJob resource to perform the deletions
 
 - Workers will then continuously pick up items from the queue and perform the actual object deletion.
 
+- We should use the single label only as there exists only AND based operator, so if we will be applying AND operator to select the resources based on the labels we need to include both `kyverno.io/ttl` and `kyverno.io/expires` as the labels in the resoource definition which is not we wanted, also there is no alternative to enable the OR based operator to select the resources to be deleted based on either of the two labels, and to enable OR based selection of resources based on the labels  we need to use the two informer factories and two set of controllers will be defined for each of the resources which will load a lot of burden on the system.
+
+- Also the absolute date format as defined in the above proposal is invalid and the default validation webhook of the cluster will throw an error so instead of `2022-08-04T00:30:00Z` as a label format, we should use `2022-08-04T003000Z` as a label format and we will handle the parsing of the label by defining a custom layout to parse this date.
+
 
 
 ## Pros And Cons of Both implementations
--- a/proposals/cleanup.md
+++ b/proposals/cleanup.md
@@ -238,6 +238,9 @@ Kyverno could alternatively leverage a CronJob resource to perform the deletions
 
 - Used auth package to check whether the service account associated with resource has permissions to get, list and delete the particular resource.
 
+- Implement the logic for listening for the creation and deletion of the specific controllers when the particular resource is added or removed from the cluster and when there is the change in the RBAC permissions for the service account associated.
+
+- Implement the logic for graceful shutdown of the controller.
 
 
 ## Pros And Cons of Both implementations
--- a/proposals/cleanup.md
+++ b/proposals/cleanup.md
@@ -47,8 +47,7 @@ There are two main use cases associated with this.
 In this proposal, Kyverno may function as a clean-up controller allowing it to identify and scavenge resources. There are two proposed capabilities which would be used to effect a clean-up.
 
 1. Use of one specific labels which may be set on any resource:
-   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`
-   - `cleanup.kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `cleanup.kyverno.io/expires: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/expires: 2022-09-30` 
+   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`, this label will also support the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 stadards. Ex., `cleanup.kyverno.io/ttl: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/ttl: 2022-09-30` 
 2. Use of a new rule type tentatively called `cleanup` which will execute in background mode on a schedule given in cron format. The rule will reuse existing Kyverno policy expression paradigms and capabilities to minimize introduced complexity.
 
 The functional requirements of both capabilities are defined below.
@@ -61,12 +60,10 @@ The functional requirements of both capabilities are defined below.
         iii. Must allow alteration after the fact
         iv. Must allow deletion to cancel the future clean-up action
         v. Must provide mechanism by which cluster operator (Kyverno CLI) can query across a cluster and return all resources which will be cleaned up
-    b. cleanup.kyverno.io/expires
-        i. Must support absolute time in future in at least one standardized time format (ISO 8601 recommended)
-        ii. Must allow both assignment to new (incoming) resources as well as assignment to existing resources
-        iii. Must allow alteration after the fact
-        iv. Must allow deletion to cancel the future clean-up action
-        v. Must provide mechanism by which cluster operator (Kyverno CLI) can query across a cluster and return all resources which will be cleaned up
+        vi. Must support absolute time in future in at least one standardized time format (ISO 8601 recommended)
+        vii. Must allow both assignment to new (incoming) resources as well as assignment to existing resources
+        viii. Must allow deletion to cancel the future clean-up action
+        ix. Must allow alteration after the fact 
     c. General:
         i. Must log clean-up of any and all resources in, minimally, the Kyverno log
         ii. Must support clean-up of both Namespaced and cluster-scoped resources (including custom)
@@ -218,41 +215,50 @@ Kyverno could alternatively leverage a CronJob resource to perform the deletions
 
 ## Implementation Approach for Label Based Resource Removal in CleanUp Policy
 
-- Design and implement metadata informer which will sync with the informer factory and will extract the labels and check for the appropriate labels.
+### Metadata Informer Design and Implementation
 
-- Then use the work queue to put the resource with the particular label in the queue and implement the business logic to schedule the deletion of the resources with the particular label schedule.
 
-- As a part of business logic, the controller will be notified as soon as the object is labeled with our ttl or expires label.
+The goal is to design and implement a metadata informer that syncs with the informer factory. This informer focuses on extracting labels and checking for appropriate labels. Specifically, it searches for the cleanup.kyverno.io/ttl label and processes resources associated with it. The following steps outline the process:
 
-- We will calculate the delay between the current time and the time object must be deleted.
+- The informer factory is established to monitor and sync resources.
 
-- We need to put the necessary information to perform the deletion in the work queue and make sure that the item should be visible in the queue when the delay calculated has elapsed.
+- The informer identifies resources with the cleanup.kyverno.io/ttl label and adds them to the factory for further processing.
 
-- Workers will then continuously pick up items from the queue and perform the actual object deletion.
+- A work queue is utilized to manage resources with this label.
 
-- We should use the single label only as there exists only AND based operator, so if we will be applying AND operator to select the resources based on the labels we need to include both `cleanup.kyverno.io/ttl` and `cleanup.kyverno.io/expires` as the labels in the resoource definition which is not we wanted, also there is no alternative to enable the OR based operator to select the resources to be deleted based on either of the two labels, and to enable OR based selection of resources based on the labels  we need to use the two informer factories and two set of controllers will be defined for each of the resources which will load a lot of burden on the system.
+- The business logic schedules the deletion of resources labeled with a specific schedule. First of all the time is parsed into the standard format and then the timestamp is added to the current time and a countdown timer is setted up after which the deletion of the resource takes place via the client.
 
-- Also the absolute date format as defined in the above proposal is invalid and the default validation webhook of the cluster will throw an error so instead of `2022-08-04T00:30:00Z` as a label format, we should use `2022-08-04T003000Z` as a label format and we will handle the parsing of the label by defining a custom layout to parse this date.
+- Upon labeling, the controller is notified and verifies the label's valid format for parsing.
 
-- Add the discovery api support which will automatically discover the resources available in the cluster and will also check that whether the resource are supporting get, list and delete verbs and also handled the skipping of the resources which do not support the listed verbs.
+- The delay between the current time and the scheduled deletion time is calculated.
 
-- Used auth package to check whether the service account associated with resource has permissions to get, list and delete the particular resource.
+- Relevant information for deletion is added to the work queue, timed to be visible when the calculated delay elapses.
 
-- Implement the logic for listening for the creation and deletion of the specific controllers when the particular resource is added or removed from the cluster and when there is the change in the RBAC permissions for the service account associated.
+- Workers retrieve items from the queue and execute actual deletions.
 
-- Implement the logic for graceful shutdown of the controller.
+### Key Points
 
-- Made the custom informers so as to support the graceful shutdown of the individual controllers and also avoiding the resource leak.
+- **Single Label**: Due to the limitation of only AND-based operator, a single label (cleanup.kyverno.io/ttl) is used to identify resources for deletion.
 
-- Implmented the graceful shutdown of the workers by using the cancellable context.
+- **Date Format**: The label's date format is adjusted from 2022-08-04T00:30:00Z to 2022-08-04T003000Z to comply with validation webhooks.
 
-- Added the unit tests for the calculation and correect parsing of the time.
+- **Informer Factory**: Custom informers ensure graceful controller shutdown and prevent resource leaks.
 
-- Added the Kuttl tests to make sure that each of the scenarios is covered gracefully by the controller.
+- **Discovery API**: A discovery API identifies resources supporting necessary verbs (get, list, delete) and skips non-compliant resources.
 
-- Added the new validation-webhook to make sure that the resources with the label value follows the specific pattern, if not a warning is thrown to the user that the label to which the Kyverno is keeping a watch doesn't follow the recommneded format.
+- **Permissions**: The auth package validates service account permissions for watch, list, and delete operations.
 
-- Created a webhook which will only keep a watch for the resources with the label `cleanup.kyverno.io/ttl` (used the labelSelector from the from the apimachinery/meta/v1 library to do so) and verfies that the label value adheres to the format defined or not.
+- **Controller Management**: Logic tracks resource creation, deletion, and changes in RBAC permissions, managing individual controllers.
+
+- **Graceful Shutdown**: Cancellable context is utilized for controllers and workers to ensure graceful shutdown.
+
+- **Unit Tests**: Tests cover time calculation and parsing to ensure accuracy.
+
+- **Kuttl Tests**: Comprehensive Kuttl tests cover various scenarios and controller behavior.
+
+- **Validation-Webhook**: A new validation webhook ensures that label values conform to the recommended format.
+
+- **Label Watcher**: A webhook monitors resources with the cleanup.kyverno.io/ttl label, leveraging labelSelector for filtering.
 
 
 ## Pros And Cons of Both implementations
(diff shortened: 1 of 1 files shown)


proposals/cleanup.md
```diff
@@ -46,15 +46,15 @@ There are two main use cases associated with this.
 
 In this proposal, Kyverno may function as a clean-up controller allowing it to identify and scavenge resources. There are two proposed capabilities which would be used to effect a clean-up.
 
-1. Use of two specific annotations which may be set on any resource:
-   - `kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the annotated resource was either created or when the annotation was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `kyverno.io/ttl: 30m`, `kyverno.io/ttl: 30d`
+1. Use of two specific labels which may be set on any resource:
+   - `kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `kyverno.io/ttl: 30m`, `kyverno.io/ttl: 30d`
```
eddycharly, line 50: I would prefer something like `cleanup.kyverno.io/ttl`

proposals/cleanup.md
```diff
@@ -46,15 +46,15 @@ There are two main use cases associated with this.
...
    - `kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `kyverno.io/expires: 2022-08-04T00:30:00Z`, `kyverno.io/expires: 2022-09-30`
```
eddycharly, line 51: I would prefer something like `cleanup.kyverno.io/expires`

proposals/cleanup.md
eddycharly: > or when the label was assigned

I don't think we can do that, it should always be relative to the resource creation time.

VedRatan: done

eddycharly:
@VedRatan We need to make sure we can watch on two labels combined with an OR.

VedRatan:
oh sure

proposals/cleanup.md
```diff
@@ -46,15 +46,15 @@ There are two main use cases associated with this.
...
+   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`
+   - `cleanup.kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `cleanup.kyverno.io/expires: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/expires: 2022-09-30` 
 2. Use of a new rule type tentatively called `cleanup` which will execute in background mode on a schedule given in cron format. The rule will reuse existing Kyverno policy expression paradigms and capabilities to minimize introduced complexity.
 
 The functional requirements of both capabilities are defined below.
 
 ```
-1. Annotations:
+1. Labels:
     a. kyverno.io/ttl
```
eddycharly, line 58: Please change the label names in the whole doc ;)

VedRatan: Sorry, didn't noticed that one.

VedRatan:
> If not possible we will have to use a single label (containing either a duration or a full date/time).

I need some time to figure out the right decision on both implementation strategies.

eddycharly:
It’s always combined with AND https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

VedRatan:
Oh, got it.

VedRatan:
Any more changes required?

eddycharly:
We’ll see during implementation.

chipzoller:
What is the plan to make this feature stateless so that failure of one of the cleanup replicas does not cause disruption?

eddycharly:
> What is the plan to make this feature stateless so that failure of one of the cleanup replicas does not cause disruption?

What do you mean ? Can you elaborate ?

chipzoller:
Assuming this reconciliation will be handled by the cleanup controller, it allows multiple replicas. If there are two replicas and the leader fails, will the other replica be able to handle all the cleanup duties for these labeled resources without losing some?

eddycharly:
Yes

chipzoller:
Ok. It's probably worth having a statement in this KDP to acknowledge this as a requirement: The solution must allow for proper tracking and accurate cleanup even upon controller replica failure or failover.

eddycharly:
The desired state can be established at any point in time from the observed state. Reconciliation should always be possible and accurate.

chipzoller:
Yes, but this needs to be explicitly tested to ensure such a failure can be tolerated. For example, if a replica failure occurs just before a resource is scheduled to be cleaned up and recovery or take-over by another replica should, within a minimal amount of time, see that resource has "missed" its cleanup interval and should be removed.

eddycharly:
Such a test sounds difficult to implement, however the design will guarantee that if a resource failed to be deleted it will be picked up by the next leader.

chipzoller:
Just saying this should be tested during development to ensure it actually happens as intended. We can handle automated tests at a later point.

proposals/cleanup.md
```diff
@@ -46,22 +46,22 @@ There are two main use cases associated with this.
 
 In this proposal, Kyverno may function as a clean-up controller allowing it to identify and scavenge resources. There are two proposed capabilities which would be used to effect a clean-up.
 
-1. Use of two specific annotations which may be set on any resource:
-   - `kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the annotated resource was either created or when the annotation was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `kyverno.io/ttl: 30m`, `kyverno.io/ttl: 30d`
-   - `kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `kyverno.io/expires: 2022-08-04T00:30:00Z`, `kyverno.io/expires: 2022-09-30` 
+1. Use of two specific labels which may be set on any resource:
```
eddycharly, line 49:
```suggestion
1. Use of one specific labels which may be set on any resource:
```

proposals/cleanup.md
```diff
@@ -46,22 +46,22 @@ There are two main use cases associated with this.
...
+   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`
+   - `cleanup.kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `cleanup.kyverno.io/expires: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/expires: 2022-09-30`
```
eddycharly, line 51: Only one label left.

proposals/cleanup.md
```diff
@@ -209,12 +209,52 @@ A couple implementations have been suggested for the annotation/label methodolog
...
+
+- Added the unit tests for the calculation and correect parsing of the time.
+
+- Added the Kuttl tests to make sure that each of the scenarios is covered gracefully by the controller.
+
+- Added the new validation-webhook to make sure that the resources with the label value follows the specific pattern, if not a warning is thrown to the user that the label to which the Kyverno is keeping a watch doesn't follow the recommneded format.
+
+- Created a webhook which will only keep a watch for the resources with the label `cleanup.kyverno.io/ttl` (used the labelSelector from the from the apimachinery/meta/v1 library to do so) and verfies that the label value adheres to the format defined or not.
+
+
```
eddycharly, line 257: Please format this section better.
Things that are important, the label, the date format, the fact that we use only metadata, permissions checks, the manager that starts/stops individual controllers, etc...

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
-   - `kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the annotated resource was either created or when the annotation was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `kyverno.io/ttl: 30m`, `kyverno.io/ttl: 30d`
-   - `kyverno.io/expires`: Sets the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 standards. Ex., `kyverno.io/expires: 2022-08-04T00:30:00Z`, `kyverno.io/expires: 2022-09-30` 
+1. Use of one specific labels which may be set on any resource:
+   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`, this label will also support the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 stadards. Ex., `cleanup.kyverno.io/ttl: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/ttl: 2022-09-30`
```
eddycharly, line 50:
```suggestion
   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the resource was created. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`, this label will also support the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 stadards. Ex., `cleanup.kyverno.io/ttl: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/ttl: 2022-09-30` 
```

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
+1. Use of one specific labels which may be set on any resource:
+   - `cleanup.kyverno.io/ttl`: Sets the time-to-live a resource may have which functions as a count-down timer starting from the time at which the labelled resource was either created or when the label was assigned. Value is a string and units are given in either minutes, hours, or days. Ex., `cleanup.kyverno.io/ttl: 30m`, `cleanup.kyverno.io/ttl: 30d`, this label will also support the absolute date and time at which clean-up should proceed. Must conform to ISO 8601 stadards. Ex., `cleanup.kyverno.io/ttl: 2022-08-04T00:30:00Z`, `cleanup.kyverno.io/ttl: 2022-09-30` 
 2. Use of a new rule type tentatively called `cleanup` which will execute in background mode on a schedule given in cron format. The rule will reuse existing Kyverno policy expression paradigms and capabilities to minimize introduced complexity.
 
 The functional requirements of both capabilities are defined below.
 
 ```
-1. Annotations:
-    a. kyverno.io/ttl
+1. Labels:
```
eddycharly, line 56:
```suggestion
1. Label:
```

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
-        i. Must support absolute time in future in at least one standardized time format (ISO 8601 recommended)
-        ii. Must allow both assignment to new (incoming) resources as well as assignment to existing resources
-        iii. Must allow alteration after the fact
-        iv. Must allow deletion to cancel the future clean-up action
-        v. Must provide mechanism by which cluster operator (Kyverno CLI) can query across a cluster and return all resources which will be cleaned up
+        vi. Must support absolute time in future in at least one standardized time format (ISO 8601 recommended)
+        vii. Must allow both assignment to new (incoming) resources as well as assignment to existing resources
+        viii. Must allow deletion to cancel the future clean-up action
+        ix. Must allow alteration after the fact 
     c. General:
```
eddycharly, line 67:
```suggestion
    b. General:
```

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
         i. Must log clean-up of any and all resources in, minimally, the Kyverno log
         ii. Must support clean-up of both Namespaced and cluster-scoped resources (including custom)
-        iii. Must only pay attention to these defined annotations (i.e., user cannot define a custom annotation as a replacement for these built-in, special Kyverno annotations)
-        iv. If both ttl and expires are specified, either Kyverno ignores the resource or it operates on the sooner of the two. Validate policy may be written which alternatively prevents this condition (account for both creation and updating a resource to add the second annotation)
```
eddycharly, line 74: NA

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
+        iii. Must only pay attention to these defined labels (i.e., user cannot define a custom label as a replacement for these built-in, special Kyverno labels)
```
eddycharly, line 70: Only one label is used now

proposals/cleanup.md
```diff
@@ -46,39 +46,36 @@ There are two main use cases associated with this.
...
+        iv. If both ttl and expires are specified, either Kyverno ignores the resource or it operates on the sooner of the two. Validate policy may be written which alternatively prevents this condition (account for both creation and updating a resource to add the second label)
 2. Rule:
     a. Must support definition in both Policy and ClusterPolicy resources.
     b. Must support a cron-style schedule around which field validation must be placed (ex., ensuring it uses only a desired subset of the total cron functionality; ensuring it does not run too frequently; etc.)
     c. Must only work in background mode and therefore must ensure background=true for this rule type
     d. Must use existing Kyverno match/exclude and other policy authoring mechanisms to define logic
     e. Must support clean-up of both Namespaced and cluster-scoped resources (including custom)
-    f. Must only clean up resources which have neither ttl nor expires annotations assigned
+    f. Must only clean up resources which have either ttl or expires labels assigned
```
eddycharly, line 78: We didn't implement this, let's clarify if we want this behaviour or not.

cc @chipzoller

proposals/cleanup.md
```diff
@@ -90,7 +87,7 @@ The functional requirements of both capabilities are defined below.
 In order to solve for the need of users to introspect an existing cluster to determine which resources may be effected by either or both capability, the following CLI use cases must be solved for.
 
 ```
-CLI Solutions for Annotations Use Case
+CLI Solutions for Labels Use Case
```
eddycharly, line 90:
```suggestion
CLI Solutions for Label Use Case
```

proposals/cleanup.md
```diff
@@ -186,14 +183,14 @@ spec:
...
-* A validate policy may be constructed preventing both the use of kyverno.io/ttl and kyverno.io/expires from being set on the same resource.
+    * Ex., write the ttl label to all Deployments in the foo Namespace
+* A validate policy may be constructed to prevent tampering with the label once a resource has been created with it assigned.
+    * Ex., block changes to or removals of ttl or expires
+* A validate policy may be constructed to prevent the use of a cleanup label based upon certain criteria.
+    * Ex., prevent the use of either label for the foo Role
+* A validate policy may be constructed preventing both the use of cleanup.kyverno.io/ttl and cleanup.kyverno.io/expires from being set on the same resource.
     * Ex., block a resource if it has both ttl and expires set
-* A validate policy may be written to enforce certain formatting/standards for these annotations
+* A validate policy may be written to enforce certain formatting/standards for these labels
```
eddycharly, line 193: Update this section to reflect the use of a single label.

proposals/cleanup.md
```diff
@@ -186,14 +183,14 @@ spec:
...
-    * Ex., block changes to or removals of ttl and/or expires
-* A validate policy may be constructed to prevent the use of a cleanup annotation based upon certain criteria.
-    * Ex., prevent the use of either annotation for the foo Role
-* A validate policy may be constructed preventing both the use of kyverno.io/ttl and kyverno.io/expires from being set on the same resource.
+    * Ex., write the ttl label to all Deployments in the foo Namespace
+* A validate policy may be constructed to prevent tampering with the label once a resource has been created with it assigned.
+    * Ex., block changes to or removals of ttl or expires
+* A validate policy may be constructed to prevent the use of a cleanup label based upon certain criteria.
+    * Ex., prevent the use of either label for the foo Role
+* A validate policy may be constructed preventing both the use of cleanup.kyverno.io/ttl and cleanup.kyverno.io/expires from being set on the same resource.
```
eddycharly, line 191: NA

proposals/cleanup.md
```diff
@@ -209,12 +206,61 @@ A couple implementations have been suggested for the annotation/label methodolog
...
 
+
+## Implementation Approach for Label Based Resource Removal in CleanUp Policy
+
+### Metadata Informer Design and Implementation
+
+
+The goal is to design and implement a metadata informer that syncs with the informer factory. This informer focuses on extracting labels and checking for appropriate labels. Specifically, it searches for the cleanup.kyverno.io/ttl label and processes resources associated with it. The following steps outline the process:
+
+- The informer factory is established to monitor and sync resources.
```
eddycharly, line 223: We don't use an informer factory anymore.

proposals/cleanup.md
```diff
@@ -209,12 +206,61 @@ A couple implementations have been suggested for the annotation/label methodolog
...
+
+- The informer identifies resources with the cleanup.kyverno.io/ttl label and adds them to the factory for further processing.
```
eddycharly, line 225: > adds them to the factory

what do you mean here ?

proposals/cleanup.md
```diff
@@ -209,12 +206,61 @@ A couple implementations have been suggested for the annotation/label methodolog
...
+
+- Workers retrieve items from the queue and execute actual deletions.
+
+### Key Points
+
+- **Single Label**: Due to the limitation of only AND-based operator, a single label (cleanup.kyverno.io/ttl) is used to identify resources for deletion.
+
+- **Date Format**: The label's date format is adjusted from 2022-08-04T00:30:00Z to 2022-08-04T003000Z to comply with validation webhooks.
+
+- **Informer Factory**: Custom informers ensure graceful controller shutdown and prevent resource leaks.
```
eddycharly, line 245: Not sure this is relevant.

proposals/cleanup.md
```diff
@@ -209,12 +206,61 @@ A couple implementations have been suggested for the annotation/label methodolog
...
+
+- **Discovery API**: A discovery API identifies resources supporting necessary verbs (get, list, delete) and skips non-compliant resources.
+
+- **Permissions**: The auth package validates service account permissions for watch, list, and delete operations.
+
+- **Controller Management**: Logic tracks resource creation, deletion, and changes in RBAC permissions, managing individual controllers.
```
eddycharly, line 251: Add a note about the interval based reconciliation (not event based).

proposals/cleanup.md
```diff
@@ -209,12 +206,61 @@ A couple implementations have been suggested for the annotation/label methodolog
...
+
+- **Graceful Shutdown**: Cancellable context is utilized for controllers and workers to ensure graceful shutdown.
+
+- **Unit Tests**: Tests cover time calculation and parsing to ensure accuracy.
+
+- **Kuttl Tests**: Comprehensive Kuttl tests cover various scenarios and controller behavior.
+
+- **Validation-Webhook**: A new validation webhook ensures that label values conform to the recommended format.
```
eddycharly, line 259: Add a note to explain that it doesn't block resource creation but returns a warning.

proposals/cleanup.md
```diff
@@ -261,13 +307,13 @@ N/A
...
+2. Kyverno resources (policies, reports, URs, etc.) themselves should either not permit being labelled with the ttl or expires label or, alternatively, Kyverno should ignore such labels on its own resources to prevent them from being cleaned up.
 
 3. Because cleanup rules may naturally require scanning the entire cluster, care should be taken to ensure, for large to huge clusters, this does not overwhelm Kyverno and result in excess processing which may cause failures in other areas. It may be prudent for such cleanup rules, which should run in background only mode, to be processed in chunks and at lower internal priority.
 
-4. If a Kyverno generate rule is written which uses a `data` type and a user specifies it should also write either the ttl or expires annotation, the user must set synchronize: false or else Kyverno must block the rule from being created.
+4. If a Kyverno generate rule is written which uses a `data` type and a user specifies it should also write either the ttl or expires label, the user must set synchronize: false or else Kyverno must block the rule from being created.
```
eddycharly, line 316:
```suggestion
4. If a Kyverno generate rule is written which uses a `data` type and a user specifies it should also write the ttl label, the user must set synchronize: false or else Kyverno must block the rule from being created.
```

Not sure i understand this point well. cc @chipzoller

proposals/cleanup.md
```diff
@@ -261,13 +307,13 @@ N/A
...
 
-1. How should Pods be handled if either the annotation is set directly on them (and not its controller) or if the rule is written which targets Pods? For the rule, should it only clean-up Pods if they do not specify an owner? Or should we prevent this rule from being created in the first place? Should we leave this as docs guidance instead?
+1. How should Pods be handled if either the label is set directly on them (and not its controller) or if the rule is written which targets Pods? For the rule, should it only clean-up Pods if they do not specify an owner? Or should we prevent this rule from being created in the first place? Should we leave this as docs guidance instead?
 
-2. Kyverno resources (policies, reports, URs, etc.) themselves should either not permit being annotated with the ttl or expires annotation or, alternatively, Kyverno should ignore such annotations on its own resources to prevent them from being cleaned up.
+2. Kyverno resources (policies, reports, URs, etc.) themselves should either not permit being labelled with the ttl or expires label or, alternatively, Kyverno should ignore such labels on its own resources to prevent them from being cleaned up.
```
eddycharly, line 312:
```suggestion
2. Kyverno resources (policies, reports, URs, etc.) themselves should either not permit being labelled with the ttl label or, alternatively, Kyverno should ignore such labels on its own resources to prevent them from being cleaned up.
```

proposals/cleanup.md
VedRatan: We have the same logic for ttl label? Am I right? If the ttl label is defined then only the resource will be removed otherwise there will be no action on the resource.

proposals/cleanup.md
VedRatan: same issue.

proposals/cleanup.md
VedRatan: sorry, I mean the work queue.

eddycharly (approved):
LGTM

JimBugwadia (approved):
Looks good overall! 

What do we do for reporting, events, metrics, etc.?  I am fine if we tackle these separately, but will be good to have some view on the observability aspects.

eddycharly:
@JimBugwadia 

We don't emit events, there's no policy to attach events to, do you think we should/can emit events ?

For metrics, we will have standard controller metrics and we will create additional metrics next week.

We also have logs when a resource is deleted, nothing more for reporting.


Anything you want to add ?

proposals/cleanup.md
realshuting: Thought we don't support expires labels?

VedRatan: Yes, you're right @realshuting updated the docs.

PR #46, opened 2023-07-27
Proposal for generating VAPs from Kyverno policies

MariamFahmy98:
Adding an initial proposal for generating validating admission policies.

### DIFF
--- a/proposals/generate_validating_admission_policies.md
+++ b/proposals/generate_validating_admission_policies.md
@@ -1,0 +1,476 @@
+# Generate Validating Admission Policies
+
+- **Authors**: [Mariam Fahmy](https://github.com/MariamFahmy98)
+- **Created**: July 18th, 2023
+- **Abstract**: Generate validating admission policies from Kyverno `validate.cel` subrule.
+
+## Overview
+- [Introduction](#introduction)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+   + [Kyverno engine VS API server](#kyverno-engine-vs-api-server)
+   + [Limitations](#limitations)
+- [Implementation](#implementation)
+   + [ClusterPolicy vs Policy](#clusterpolicy-vs-policy)
+   + [Kyverno Policy Settings](#kyverno-policy-settings)
+   + [Match/Exclude Resources](#matchexclude-resources)
+   + [Validate Rules](#validate-rules)
+   + [Audit Annotations](#audit-annotations)
+   + [Parameter Resources](#parameter-resources)
+   + [Limitations](#limitations-1)
+- [Next Steps](#next-steps)
+
+## Introduction
+Kyverno policies support CEL expressions for resources validation. The goal is to generate Kubernetes validating admission policies and validating admission binding from the information provided by Kyverno policies.
+
+- In case a new Kyverno policy is created, a validating admission policy and its binding are generated. 
+- In case Kyverno policy is updated, the corresponding VAP and its binding needs to be updated accordingly. 
+- In case Kyverno policy is deleted, the corresponding VAP and its binding are deleted.
+
+Once validating admission policy and its binding are generated, the API server will handle the resource validation. 
+
+An example of Kyverno policy from which validating admission policy and its binding will be generated:
+```yaml=
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: check-deployment-replicas
+spec:
+  validationFailureAction: Enforce
+  background: false
+  rules:
+    - name: deployment-replicas
+      match:
+        any:
+        - resources:
+            kinds:
+              - Deployment
+      validate:
+        cel:
+          expressions:
+            - expression: "object.spec.replicas < 5"
+              message:  "Deployment spec.replicas must be less than 5”
+```
+
+## Motivation
+Currently, Kyverno engine supports CEL expressions for resource validation. However, validating admission policy is an alternative solution to validating admission webhooks. Therefore, it's better to make Kyverno generate validating admission policies so that resource validation will be handled by the API server instead. Moreover, the resource validation aganist policies won't be affected in case Kyverno isn't available for some reason.
+
+Users can either let Kyverno engine to validate resources using CEL expressions or let the API server to validate resources and this is done by allowing Kyverno to generate validating admission policies and their bindings from a Kyverno policy containing a CEL subrule.
+
+## Proposal
+- A new controller `validatingadmissionpolicy-controller` will be created. It will watch the following:
+  1. Kyverno cluster policies
+  2. Validating admission policies
+  3. Validating admission policy bindings
+
+- When a policy event happens, the policy key is enqueued. The controller will create/update/delete validating admission policies and their bindings accordingly.
+
+- When a validating admission policy or validating admission policy binding event happens, the policy key is deduced from owner references and it's enqueued. The controller will reset the changes done in either VAP or VAP binding.
+
+- A flag/toggle `--generateValidatingAdmissionPolicy` will be introduced in the controller so that user can either let Kyverno engine handles CEL expressions or let api server handles it by generating VAPs and their bindings. By default, it is set to false.
+
+### Kyverno engine VS API server
+Since CEL expressions can be handled by either Kyverno engine or the API server, Here are the cases where Kyverno engine will handle the resource validations itself:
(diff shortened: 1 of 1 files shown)


proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+              - Deployment
+      validate:
+        cel:
+          expressions:
+            - expression: "object.spec.replicas < 5"
+              message:  "Deployment spec.replicas must be less than 5”
+```
+
+## Motivation
+Currently, Kyverno engine supports CEL expressions for resource validation. However, validating admission policy is an alternative solution to validating admission webhooks. Therefore, it's better to make Kyverno generate validating admission policies so that resource validation will be handled by the API server instead.
```
JimBugwadia, line 56: maybe elaborate on why its better to execute a policy in the API server?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+
+Users can either let Kyverno engine to validate resources using CEL expressions or let the API server to validate resources and this is done by allowing Kyverno to generate validating admission policies and their bindings from a Kyverno policy containing a CEL subrule.
+
+## Proposal
+- A new controller `validatingadmissionpolicy-controller` will be created. It will watch the following:
+  1. Kyverno policies
```
JimBugwadia, line 62: all, or with a specific label?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+- When a policy event happens, the policy key is enqueued. The controller will create/update/delete validating admission policies and their bindings accordingly.
+
+- When a validating admission policy or validating admission policy binding event happens, the policy key is deduced from owner references and it's enqueued. The controller will reset the changes done in either VAP or VAP binding.
+
+- A flag/toggle will be introduced in the controller so that user can either let Kyverno engine handles CEL expressions or let api server handles it by generating VAPs and their bindings.
+
+### Kyverno engine VS API server
+Since CEL expressions can be handled by either Kyverno engine or the API server, Here are the cases where Kyverno engine will handle the resource validations itself:
+1. The flag is disabled in the `validatingadmissionpolicy-controller`. This is done by setting `--generateValidatingAdmissionPolicy` to false.
+2. The flag is enabled and validating admission policies were generated but issues are encountered in processing these policies. This is done by getting the metrics exposed by the API server.
```
JimBugwadia, line 75: are there additional details on what exactly we can check to see if a policy has been executed in the API server or not?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+
+### Limitations
+Validating admission policy and its bindings will be generated for a Kyverno policy that has one `validate.CEL` subrule.
```
JimBugwadia, line 78: one or at least one?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+
+## Implementation
+In this section, we will discuss how validating admission policy and its binding can be created from Kyverno policies.
+
+### ClusterPolicy vs Policy
+All Kyverno policies of kind `ClusterPolicy` can be converted into validating admission policies but not all policies of kind `Policy` can be converted into validating admission policies since they are cluster-scoped resources.
```
JimBugwadia, line 84: What about a `ClusterPolicy` with API calls, external service calls, pre-conditions, etc. 

Those cannot be converted, right?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+            - prod
+```
+
+### Kyverno Policy Settings
+In Kyverno policies, some common settings can be applied to all rules. We will discuss the effect of setting these values on generating validating admission policies assuming that Kyverno policy contains multiple rules, one of them is `validate.CEL` subrule.
+
+1. `applyRules`: 
+    - If it's set to `All` (default), Validating admission policy and its binding will be generated as a result of having `validate.CEL` subrule. 
+    - If it's set to `One` and `validate.CEL` subrule is the first matching rule, then Validating admission policy and its binding will be generated.
+    - If it's set to `One` and `validate.CEL` subrule isn't the first matching rule, no validating admission policy will be generated.
```
JimBugwadia, line 117: I don't think this will provide the intended behaviors.

Rules are an ordered set in a Kyverno policy, and `applyRules` controls when to stop processing of rules.

It may be best to limit the generation of Validating Admission Policies to policies that contain a single rule, for the initial release.

JimBugwadia (commented):
Added a few comments.

Will we also indicate in the Policy Status if the policy resulted in a generated Validating Admission Policy? A `validatingAdmissionPolicy` element in the status may be useful to show if this was done and also record any errors, etc.

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+
+## Proposal
+- A new controller `validatingadmissionpolicy-controller` will be created. It will watch the following:
+  1. Kyverno policies
+  2. Validating admission policies
+  3. Validating admission policy bindings
+
+- When a policy event happens, the policy key is enqueued. The controller will create/update/delete validating admission policies and their bindings accordingly.
+
+- When a validating admission policy or validating admission policy binding event happens, the policy key is deduced from owner references and it's enqueued. The controller will reset the changes done in either VAP or VAP binding.
```
realshuting, line 68: Is VAP (binding) a namespaced resource?

proposals/generate_validating_admission_policies.md
```diff
@@ -0,0 +1,474 @@
...
+
+- A flag/toggle will be introduced in the controller so that user can either let Kyverno engine handles CEL expressions or let api server handles it by generating VAPs and their bindings.
```
realshuting, line 70: The default is on (generate VAPs and bindings)?

chipzoller:
Would like to review this once above comments have been addressed and resolved.

proposals/generate_validating_admission_policies.md
MariamFahmy98: If the controller is enabled, then it will watch all Kyverno policies and check if it's `validate.CEL` subrule.

proposals/generate_validating_admission_policies.md
MariamFahmy98: one `validate.CEL` subrule as we will focus on generating VAPs for a policy that has one rule for now.

proposals/generate_validating_admission_policies.md
MariamFahmy98: Yes that's correct.

proposals/generate_validating_admission_policies.md
MariamFahmy98: No, they are cluster scoped resources.

proposals/generate_validating_admission_policies.md
MariamFahmy98: The default is off because clusters with versions lower than 1.26 don't have validating admission policies so I make it off as default and if the user has the prerequisites to use VAPs then he can enable it.

The minimum requirements are:
1. Kubernetes >= v1.26
2. ValidatingAdmissionPolicy feature gate is enabled.
3. admissionregistration.k8s.io/v1alpha1 API is enabled.

proposals/generate_validating_admission_policies.md
realshuting: What if a user defines a sub-VAP rule in a Policy? The namespaced resource cannot be the owner of a cluster-wide resource.

proposals/generate_validating_admission_policies.md
MariamFahmy98: Upon investigation, we have the following metrics:
1. apiserver_validating_admission_policy_definition_total: Validation admission policy count total, labeled by state and enforcement action.
2. apiserver_validating_admission_policy_check_total: Validation admission policy check total, labeled by policy and further identified by binding, enforcement action taken, and state.
3. apiserver_validating_admission_policy_check_duration_seconds: Validation admission latency for individual validation expressions in seconds, labeled by policy and further including binding, state and enforcement action taken.

There are no metrics to check the VAP execution in the API server.

chipzoller (approved):

realshuting (approved):

PR #47, opened 2023-08-08
global context proposal

JimBugwadia:


MariamFahmy98:
I can't find the resource cache proposal in the KDP. 
Am I missing something?

JimBugwadia:
> I can't find the resource cache proposal in the KDP. Am I missing something?

Oops...forgot to add the file! 

As they say, good design is invisible :-) 

Added.

README.md
```diff
@@ -23,6 +23,8 @@ Older proposals were managed in documents. All new proposals should be submitted
 | [Store Kyverno policies in OCI registries](https://docs.google.com/document/d/15cqD4HPecI5Uv2u1Yfg0JCgWDVi2HLwGZPvTX_48W2E/edit?usp=sharing) | Implemented | 1.9 |
 | [Policy Exceptions](https://github.com/kyverno/KDP/pull/33) | In Review | 1.9 |
 | [ConfigMap cache enhancement with Informers](/proposals/cache_enhancements.md) | Implemented | 1.9 |
+| [Resource Cache](https://github.com/kyverno/KDP/pull/47) | Proposal | 1.11 |
```
chipzoller, line 26: If we're going to maintain a list of proposed and in-review KDPs in the README file which reference PRs, then the update to the README probably needs to be a separate PR so it gets merged first and acts as a ToC to other PRs (some of which may not ever be accepted).

proposals/resource_cache.md
```diff
@@ -0,0 +1,171 @@
...
+```yaml
+      context:
+        - name: hosts
+          apiCall:
+            urlPath: "/apis/networking.k8s.io/v1/ingresses"
+            jmesPath: "items[].spec.rules[].host"
+            cache: true
+```
+
+# Implementation
```
chipzoller, line 56: Don't see it addressed here, but with `cache: true` will Kyverno _ONLY_ return resources specified by that apiCall AND those which have been successfully cached (have `cache.kyverno.io/enabled: "true"`)? Ex., the call for Ingress resources is used per above. 20 Ingress resources exist but 0 have been labeled. Are exactly 0 returned? If so, does the call fail because of 0 or succeed with an empty set?

What about if there are still 20 total but 10 have been labeled? Will 10 or 20 be returned?

proposals/resource_cache.md
```diff
@@ -0,0 +1,171 @@
...
+
+When policies are created or modified, Kyverno will attempt to initialize informers for any resource type when `cache: true` is specified in the `apiCall`. In case an informer cannot be initialized, or the `urlPath` cannot be converted to an cache lookup (see [Converting API Calls](#converting-api-calls), an error will be returned.
```
chipzoller, line 58: What about when rules are added and deleted? I would expect informers to be created and destroyed, respectively.

chipzoller (commented):
It seems like it may also be a good idea to register a new metric which indicates how many total cache informers exist at a given time to assist users (and maintainers) in troubleshooting.

proposals/resource_cache.md
JimBugwadia: separated cache from API call

proposals/resource_cache.md
JimBugwadia: yes

proposals/resource_cache.md
```diff
@@ -0,0 +1,243 @@
...
+      version: "v1"
+      kind: "ingresses"
+      jmesPath: "ingresses | items[].spec.rules[].host"
+```
+
+This allows policy authors to declare what resource should be cached. 
+
+The `group` and `version` are optional. If not specified, the preferred versions should be used.
+
+An optional `namespace` can be used to only cache resources in the namespace, rather than across all namespaces which is the behavior is a namespace is not specified.
```
realshuting, line 60: Would be good to add it to the example snippet.

eddycharly:
Ooops, I missed that, will look at it next week !

proposals/resource_cache.md
chipzoller: Resolved?

chipzoller (changes requested):
What will happen if a resourceCache context variable is declared yet there are no corresponding resources in the cache? Does the context var fail to be evaluated, does it evaluate automatically to `[]`, something else?

proposals/resource_cache.md
JimBugwadia: Yes.

proposals/resource_cache.md
```diff
@@ -0,0 +1,271 @@
...
+        kind: "networkpolicies"
+        namespace: "{{request.namespace}}"
+        jmesPath: "items[?label_match(spec.podSelector.matchLabels, `{{request.object.spec.template.metadata.labels}}`)] | length(@)"
+```
+
+# Drawbacks
+
+API calls do not leverage caching by default.
+
+If needed, we can add a separate caching mechanism for API calls in the future.
```
vishal-chdhry, line 234: Updating drawbacks of this approach

```suggestion
1. API calls do not leverage caching by default. If needed, we can add a separate caching mechanism for API calls in the future.
2. Using `cache.kyverno.io/enabled: "true"` can cause issues when users forgets to add them to the resource. 
   1. It is easy to miss a resource when adding the label. `apiCall` will return all the resources of the given type while `resourceCache` will return only those resources that have the label. In case of 1-10k resources. 
   2. Users might not want to have a kyverno specific label in all their resources across all namespaces.
3. For the usecase mentioned in [motivation](#motivation), we need to add resource to the cache when the policies is applied, otherwise, when the resource is applied for the first time, it will fail because of the timeout like it currently does. This will take away the abilities to have substitutions in `resourceCache` (e.g. `namespace: "{{request.namespace}}"`) and the `resourceCache` field will have to be static.
```

proposals/resource_cache.md
```diff
@@ -0,0 +1,263 @@
...
+
+There are two parts to this feature:
+1. Allow users to manage what data should be loaded in a global context.
+2. Allow policy rules to reference global context data.
+
+The global context is populated when Kyverno runs, and can then be referenced by multiple policy rules. Kubernetes resource data, fetched from the API server is automatically updated based on events. Data fetched from other sources can be periodically updated.
+
+Users can manage which resources to cache by creating a new custom resource called `GlobalContextEntry` provided by Kyverno. This will decouple the creation and usage of a global entry. 
+
+A `GlobalContextEntry` will can be either of the following types:
```
realshuting, line 59:
```suggestion
A `GlobalContextEntry` can be either of the following types:
```

chipzoller:
The two things I, as a user, really need answered here are:
1. what happens when I try and fetch something from global cache and it's empty?
2. what does the returned data look like when when the cache is populated? I want it to look the same as if I did a raw API call to the K8s API so I can avoid having to write separate JMESPath queries.

realshuting:
> The two things I, as a user, really need answered here are:
> 
> 1. what happens when I try and fetch something from global cache and it's empty?
> 2. what does the returned data look like when when the cache is populated? I want it to look the same as if I did a raw API call to the K8s API so I can avoid having to write separate JMESPath queries.

cc @vishal-chdhry @KhaledEmaraDev

vishal-chdhry:
1. In the original implementation, if the context entry is not present, the policy would have returned an error. Khaled can you verify that this is the current behaviour.

2. Unfortunately, there is a slight difference, when we do an API call and dont mention the resource, we get a list in return. But in the global context, we are using imformers and informer returns an array of object. Basically the output of global context is .Items of the corresponding apicall

KhaledEmaraDev:
> > The two things I, as a user, really need answered here are:
> > 
> > 1. what happens when I try and fetch something from global cache and it's empty?
> > 2. what does the returned data look like when when the cache is populated? I want it to look the same as if I did a raw API call to the K8s API so I can avoid having to write separate JMESPath queries.
> 
> cc @vishal-chdhry @KhaledEmaraDev 

1. Just like Vishal mentioned. Plus, it would set the policy to NotReady and with the correct reason of the Cache Entry not existing.

2. Again Vishal is correct here. APICall would return a DeploymentList array each of which would have its own items. So it has to be queried the old way for instance `[0].items | length(@)` while the resource would return `items` to begin with.

realshuting:
Global context is supported, see docs https://kyverno.io/docs/policy-types/cluster-policy/external-data-sources/#global-context.

realshuting (approved):

PR #53, opened 2023-12-12
add a proposal for CRD conversion webhook

MariamFahmy98:


proposals/conversion-webhook.md
```diff
@@ -0,0 +1,73 @@
...
+Some common use cases where conversion is required:
+1. custom resource is requested in a different version than stored version.
+2. Watch is created in one version but the changed object is stored in another version.
+3. custom resource PUT request is in a different version than storage version.
+
+## Motivation
+Kyverno CRDs `v1` contain many deprecated fields that require removal. Additionally, there are fields at the `spec` level that are only specific to certain rules and not others. Therefore, we would like to introduce a more stable version, `v2`. Since `v1` and `v2` will have different schema, a conversion webhook is required to convert between them.
+
+## Proposal
+In Kubernetes, it is essential that all versions are compatible with each other. This means that if we go from `v1` to `v2` and then back to `v1`, we must not lose any information. Therefore, any changes we make to our API must be compatible with `v1`, and we must ensure that anything we add in `v2` is also supported in `v1`. This requires us to add new fields to `v1` to support any new additions.
```
MarcelMue, line 26: > This requires us to add new fields to `v1` to support any new additions.

I don't quite understand this, isn't it possible for us to use an annotation (as stated in solution 2.) also on the V1 object? So if something is only available in v2, it gets written to an annotation in v1 and then we don't lose data. Maybe I am missing something why this wouldn't work.

MariamFahmy98: We can do that as well.

proposals/conversion-webhook.md
```diff
@@ -0,0 +1,73 @@
...
+   - Deprecate v2beta1 and v1.
+   - The storage version is v1.
+2. For Kyverno v1.13:
+   - The storage version will be v2.
+3. For Kyverno 1.14:
+   - v2beta1 will be removed.
+
+Note that multiple versions may exist in storage if they were written before the storage version changes – changing the storage version only affects how objects are created/updated after the change.
+
+## Implementation
```
MarcelMue, line 71: I think implementation is the most important thing to clarify here. The two standouts to me are:
1. Testing of the conversion
2. Deployment of the CRDs with the webhook included in helm (size limitation issues + templating necessity)

I think the proposal here might change slightly when implementation plans become more clear.

realshuting:
Closing as not needed.

PR #54, opened 2023-12-12
Proposal: policy based webhook configurations

realshuting:
This PR proposes the strategy to configure webhooks based on policy settings.

Related requests:
https://github.com/kyverno/kyverno/issues/9111
https://github.com/kyverno/kyverno/issues/8063


chipzoller:
Is this ready for final review? If not, consider placing in draft mode.

realshuting:
> Is this ready for final review? If not, consider placing in draft mode.

Yes, added to the matintainers meeting.

proposals/per_policy_webhook_configurations.md
```diff
@@ -0,0 +1,235 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Prior Art](#prior-art)
+- [Unresolved Questions](#unresolved-questions)
+- [CRD Changes (OPTIONAL)](#crd-changes-optional)
+
+# Overview
+[overview]: #overview
+
+Configure Kyverno resource webhook configurations based on policy settings.
```
chipzoller, line 27: This overview isn't very descriptive and should be elaborated upon. Kyverno today does a form of this and so it needs to be explained.

proposals/per_policy_webhook_configurations.md
```diff
@@ -0,0 +1,235 @@
...
+
+# Definitions
+[definitions]: #definitions
+
+Kyverno has two webhook configurations registered for the selected resources by policies, a MutatingWebhookConfiguration and a ValidatingWebhookConfiguration respectively. For brevity, this proposal simply uses "webhooks" to refer to the two mentioned webhook configurations.
```
chipzoller, line 32: Are you also only talking about the `-resource` webhook for each? There are different webhooks for different purposes.

chipzoller (commented):
Haven't done a full review, but we discussed having per-rule webhooks in the past which, although I don't see it stated, seems to be a complimentary decision. Is that not still on the table?

proposals/per_policy_webhook_configurations.md
```diff
@@ -0,0 +1,235 @@
...
+```
+The above rule will be transformed to the following webhook configuration.
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+webhooks:
+- name: my-webhook.example.com
+  objectSelector:
+    matchLabels:
+      foo: bar
```
MarcelMue, line 77: Is it possible to use set-based filtering here? (ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering)

Otherwise I think this will be messy IMO, constructing an "OR" for these two policy cases:
```
spec:
  rules:
    - match:
        all:
        - resources:
            selector:
              matchLabels:
                foo: bar
```
```
spec:
  rules:
    - match:
        all:
        - resources:
            selector:
              matchLabels:
                foo: foo
```

Would end up in two unique webhook configurations in the ValidatingWebhookConfiguration -  or alternatively only `matchConditions` should be used

MariamFahmy98: Another thing is what happens if we have a policy that matches all resources with a specific label and another one that matches a configmap for example? We will end up have a webhook configuration that matches configmaps with this label.

Example:
1.  The first policy matches all resources with a specific label:
```
spec:
  rules:
    - match:
        all:
        - resources:
            selector:
              matchLabels:
                foo: bar
```
2. The second policy matches configmaps:
```
rules:
  - match:
      any:
      - resources:
          kinds:
          - ConfigMap
```

The generated configuration will be as follows:
```
  objectSelector:
    matchLabels:
      foo: bar
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    - UPDATE
    - DELETE
    - CONNECT
    resources:
    - configmaps
   ```
Therefore, configmaps with this label will only be forwarded to the webhook server.

File: proposals/per_policy_webhook_configurations.md (hunk 2 above)
Comment (MariamFahmy98) [reply proposals/per_policy_webhook_configurations.md:32]: I believe it will be for `-resource` webhooks only.

File: proposals/per_policy_webhook_configurations.md [hunk 4]
```diff
@@ -0,0 +1,235 @@
...
+The above rule will be transformed to the following webhook configuration.
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+webhooks:
+  - name: my-webhook.example.com
+    # note: Kubernetes allows up to 64 matchConditions per webhook
+    matchConditions:
+    - name: 'exclude-kubelet-requests'
+      expression: '!("system:nodes" in request.userInfo.groups)'
```
MariamFahmy98, line 131: All webhook `matchConditions` must evaluate to true. It isn't OR-ed operation so we will have the same issue as above. 
I believe configuring webhooks `objectSelector`, `namespaceSelector` and `matchConditions` from the policy settings isn't a good idea as we will end up having multiple webhook configurations.

proposals/per_policy_webhook_configurations.md
MarcelMue: It seems to me like our best course of action is to have a hierarchy for which changes are made to the webhook config. So if e.g. some policy has `any resource` and another has `resource with label` then our logic should decide that `any resource` > `resource with label`.
Then only the rule without objectselector should be added.

It sounds easy but I expect this to be reasonably complex.

proposals/per_policy_webhook_configurations.md
realshuting: Yes, for resource webhooks only.

proposals/per_policy_webhook_configurations.md
realshuting: Will update it.

proposals/per_policy_webhook_configurations.md
realshuting: I wonder if we could start with the `matchConditions` and make the webhook rule and kyverno rule 1-1 map. This way it requires users to configure webhook filters with caution.

proposals/per_policy_webhook_configurations.md
JimBugwadia, line 32:
```suggestion
Kyverno has two webhook configurations that are auto-created and managed for resources selected by policies, a MutatingWebhookConfiguration and a ValidatingWebhookConfiguration respectively. For brevity, this proposal simply uses "webhooks" to refer to the two mentioned webhook configurations.
```

proposals/per_policy_webhook_configurations.md
JimBugwadia, line 27: Perhaps we should briefly describe the current scheme and why that is insufficient?

realshuting: Updated. 

I believe the benefits under `Motivation` section point out current issues. Does that make sense to you?

realshuting: >but we discussed having per-rule webhooks in the past which, although I don't see it stated, seems to be a complimentary decision. Is that not still on the table?

I linked to the other policy-based webhook configuration PR https://github.com/kyverno/kyverno/issues/8063 in the description. Is there any other prior issue logged for this?

JimBugwadia:
@realshuting - have we decided on the approach for handling admission requests for per-policy webhooks vs the current approach of finding all matching policies and rules for an admission request?

realshuting:
Can we merge this since the change was merged to 1.12?

JimBugwadia (approved):

MariamFahmy98 (approved):

PR #55, opened 2024-01-12
Add a proposal for generating VAPs given an exception

MariamFahmy98:
This PR includes the latest investigation report on the generation of VAP in case of a policy exception.

### DIFF
--- a/proposals/generate_validating_admission_policies.md
+++ b/proposals/generate_validating_admission_policies.md
@@ -2,6 +2,7 @@
 
 - **Authors**: [Mariam Fahmy](https://github.com/MariamFahmy98)
 - **Created**: July 18th, 2023
+- **Modified**: Dec 28th, 2023
 - **Abstract**: Generate validating admission policies from Kyverno `validate.cel` subrule.
 
 ## Overview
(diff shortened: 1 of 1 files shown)


proposals/generate_validating_admission_policies.md
```diff
@@ -468,9 +473,171 @@ spec:
...
+        resources:     ["deployments"]
+        resourceNames: ["important-tool"]
+      namespaceSelector:
+        matchLabels:
+          environment: staging
+    validations:
+      - expression: "'app' in object.metadata.labels"
+  ```
+
+  However, using a `namespaceSelector` will be applied for both matching and excluding deployemnts. The above VAP matches all deployments whose namespace labels set to `environment: staging`, and it also excludes all deployments named `important-tool` whose namespace labels set to `environment: staging`. It means that matching/excluding resources is done in the same namespace. There is no option to match deployments in all namespaces and exclude some in a specific namespace.
```
JimBugwadia, line 562: @MariamFahmy98 - isn't this the same as what this Kyverno declaration would do?

```yaml
...
  rules:
  - name: check-deployment-labels
    match:
      all:
      - resources:
          namespaceSelector:
            matchLabels:
              environments: staging
    exclude:
      all:
        - resources:
            name: important-tool
```

proposals/generate_validating_admission_policies.md
```diff
@@ -468,9 +473,171 @@ spec:
...
+      namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: delta
+  ```
+
+  However, the use of `namespaceSelector` limits resource matching to the same namespace as the exclusion.
+
+- An exception that matches resources based on annotations, subjects, roles, or clusterRoles cannot be converted to VAPs.
+
+In conclusion, exceptions provide the flexibility to exclude resources that VAPs lack, making it impossible to generate a VAP from an exception. As a result, if an exception is created for a Kyverno policy that uses `validate.cel` subrule, the Kyverno engine will handle the resource validation itself instead of generating the corresponding VAPs.
```
JimBugwadia, line 637: Can we allow use of some subset of exceptions?

proposals/generate_validating_admission_policies.md
MariamFahmy98: The corresponding VAP could be something like this:

MariamFahmy98: The corresponding VAP could be something like this:
```
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingAdmissionPolicy
metadata:
  name: "check-deployment-labels"
spec:
  failurePolicy: Fail
  matchConstraints:
    excludeResourceRules:
    - apiGroups:     ['*']
      apiVersions:   ['*']
      resources:     ['*']
      resourceNames: ["important-tool"]
    namespaceSelector:
      matchLabels:
        environment: staging
  validations:
    - expression: "'app' in object.metadata.labels"
```
But I need to double-check if this is correct.

proposals/generate_validating_admission_policies.md
MariamFahmy98: For the following cases, I think a VAP can be generated:
1. Matching all resources and excluding a specific one.
2. Matching and excluding in the same namespace.

However, there are a lot of limitations and I don't know if it is worth considering this feature.

JimBugwadia: Can we clarify the limitations and propose changes to ValidatingAdmissionPolicy resources?

https://github.com/kubernetes/kubernetes/issues/122205#issuecomment-1927390823

realshuting (approved):

realshuting:
Supported via https://github.com/kyverno/kyverno/pull/9386.

PR #62, opened 2024-08-10
proposal: Add mutation support for the kyverno CLI

aerosouund:
Proposal document that addresses https://github.com/kyverno/kyverno/issues/4354

### DIFF
--- a/proposals/mutation-support-cli.md
+++ b/proposals/mutation-support-cli.md
@@ -1,0 +1,258 @@
+# Kyverno Design Proposal - Add mutation support in the CLI
+
+Created: Aug 10th, 2024
+Author: [Ammar Yasser](https://github.com/aerosouund)
+
+- [Overview](#overview)
+  * [Online Operation](#online-operation)
+  * [Offline Operation](#offline-operation)
+- [Proposal](#proposal)
+- [Example](#example)
+- [Implementation](#implementation)
+  * [Resources obtained from a cluster](#resources-obtained-from-a-cluster)
+  * [Resources passed from the CLI directly](#resources-passed-from-the-cli-directly)
+  * [Previous attempts](#previous-attempts)
+
+<small><i><a href='http://ecotrust-canada.github.io/markdown-toc/'>Table of contents generated with markdown-toc</a></i></small>
+
+## Overview
+
+Kyverno webhook controller runs as an admission controller in Kubernetes, and it mutates resources upon admission review. Such mutation is performed based on resources operations (CREATE/UPDATE/DELETE) via webhook.
+This feature currently is included in the webhook controller only, the CLI has the current behavior:
+
+### Online Operation
+an operation performed on a live kubernetes cluster and has the `--cluster` flag passed
+
+**Behavior**: No mutation occurs on the target resources fetched from the cluster.
+The output is just how many resources passed or failed as typical in the output of the cli.
+
+`pass: 5, fail: 0, warn: 0, error: 0, skip: 0`
+
+**Technical Explanation**: The CLI succeeds in fetching the correct resources and applies the mutation properly. but what gets returned to the report generator is still the original resource
+
+`pkg/engine/mutation.go`:
+
+```go
+resource, ruleResp := e.invokeRuleHandler(
+	ctx,
+	logger,
+	handlerFactory,
+	policyContext,
+	matchedResource,
+	rule,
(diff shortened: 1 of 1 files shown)


proposals/mutation-support-cli.md
```diff
@@ -0,0 +1,325 @@
...
+resources:
+  - <path/to/resource.yaml>
+  - <path/to/resource.yaml>
+exceptions:
+  - <path/to/exception.yaml>
+  - <path/to/exception.yaml>
+targetResources: # THIS WILL BE ADDED
+  - <path/to/unmutated/resource.yaml>
+variables: variables.yaml
+mutateExistingTargetVariables: mutate-vars.yaml  # THIS WILL BE ADDED
```
realshuting, line 108: Can we reuse the existing `variables` attribute?

proposals/mutation-support-cli.md
```diff
@@ -0,0 +1,325 @@
...
+        - resources:
+            kinds:
+            - Deployment
+            namespaces:
+              - "secure"
+      mutate:
+        targets:
+        - apiVersion: apps/v1
+          kind: Deployment
+          namespace: "secure"
```
realshuting, line 159: Let's use a different example to illustrate the mutate existing policy, i.e., we can choose to mutate an existing configmap upon namespace creation. The current one can be achieved via the standard admission review mutation.

proposals/mutation-support-cli.md
```diff
@@ -0,0 +1,325 @@
...
+	
+
+## Proposal
+
+### apply
+
+The apply command will receive an additional flag `--target-resource`. this flag can be passed multiple times for single resources. e.g:
+
+```bash
+kyverno apply policy-with-mutate-existing.yaml --target-resource=resource1.yaml  --target-resource=resource2.yaml
```
realshuting, line 71: Please attach the full command, how do we accept the trigger resource?

proposals/mutation-support-cli.md
```diff
@@ -0,0 +1,393 @@
...
+- [Implementation](#implementation)
+  * [Resources obtained from a cluster](#resources-obtained-from-a-cluster)
+  * [Resources passed from the CLI directly](#resources-passed-from-the-cli-directly)
+  * [Previous attempts](#previous-attempts)
+
+<small><i><a href='http://ecotrust-canada.github.io/markdown-toc/'>Table of contents generated with markdown-toc</a></i></small>
+
+## Overview
+
+The Kyverno webhook controller runs as an admission controller in Kubernetes, mutating resources upon admission review. Such mutation is performed based on resource operations (CREATE/UPDATE/DELETE) via the webhook. This feature is currently available only in the webhook controller. The CLI has the following current behavior:
```
realshuting, line 20:
```suggestion
The Kyverno webhook controller runs as an admission controller in Kubernetes, mutating existing resources upon admission review. Such mutation is performed based on resource operations (CREATE/UPDATE/DELETE) via the webhook. This feature is currently available only in the webhook controller. The CLI has the following current behavior:
```

proposals/mutation-support-cli.md
```diff
@@ -0,0 +1,393 @@
...
+targetResources:
+  - name: nginx-demo1
+    values:
+      request.operation: CREATE
+  - name: nginx-demo2
+    values:
+      request.operation: UPDATE
+```
+
+The `targetResources` key will be added, with an array of objects containing two keys: name (a string) and values (a `map[string]interface{}`).
```
realshuting, line 135: You need an unique identifier for the target resource. Having the name only could cause the confusion if two resources are with the same name but different namespaces.

aerosouund: This is how currently resources are being identified for policy variables as well.
We can improve this by adding a `namespace` variable to this schema. This will apply to target resource variables and policy variables

realshuting: The current solution needs to be improved. Though this is not a major use case, duplicate resource names could be provided to the CLI.

aerosouund: Yeah, which is why i was kinda reluctant to consider this case. because it's already a pretty niche case and i didn't know if the complexity of the change will equal the value gained from addressing it.
I will include it and we can change it back later

realshuting (approved):

MariamFahmy98 (approved):

PR #63, opened 2024-08-12
feat: add audit mode and reporting in mutate and generate rules

vishal-chdhry:
## Summary

This proposal discusses ways to capture mutate and generate rule results in existing policy reports.


proposals/mutate-and-generate-reporting.md
```diff
@@ -0,0 +1,132 @@
...
+[motivation]: #motivation
+
+Currently, reporting is only supported in validate rules. While this is nice in that it allows a report to be generated showing the cluster resources that violate the validation rules, it doesn't show the resources that would have either been mutated or generated by Kyverno policy. As an operator, administrator, or manager who may have multiple clusters under management, this makes it impossible to know if the mutate or generate rules about to be pushed out are safe and effective.
+
+# Proposal
+
+The `properties` field in policy report result is used to add information regarding mutation and generation. A `type` property is used to indicate if the report is for mutate, validate or generate.
+For mutate rules, the policy report should indicate, minimally,:
+- jsonpatch of the mutation
+- target of the mutation
```
MariamFahmy98, line 35: The target is only applicable for mutate existing rules.

MariamFahmy98 (commented):
Few questions:
- I expect that those reports will be generated as a result of admission requests. Do we need to consider background scanning? 
  IMO background scanning for standard mutation and standard generation isn't needed because we do have mutateExisting rules as well as generateExisting rules and we can generate reports for them too. 
  
  That's why I have in mind:
   1. standard mutation rules --> admission reports
   2. standard generation        --> admission reports
   3. mutateExisting rules        --> background scan reports
   4. generateExisting rules     --> background scan reports
   
   We need to discuss the effect of setting `spec.background` and `spec.admission` in each case.

- Does the value of `validationFailureAction` has any effect on generating the reports specifically for mutation rules? Given that this field under the `spec` level is deprecated.

realshuting:
>I expect that those reports will be generated as a result of admission requests. Do we need to consider background scanning?
IMO background scanning for standard mutation and standard generation isn't needed because we do have mutateExisting rules as well as generateExisting rules and we can generate reports for them too.

>That's why I have in mind:

>standard mutation rules --> admission reports
standard generation --> admission reports
mutateExisting rules --> background scan reports
generateExisting rules --> background scan reports


Good point! The standard mutate rules are applied upon admission review, I don't think we need to refresh the reports for that during the background scan. For mutate existing and generate (regardless of generateExisting) rules, they are applied during the background scan so we do need to generate/refresh reports for these types of rules.

realshuting:
This is supported, see https://kyverno.io/blog/2024/10/30/announcing-kyverno-release-1.13/#warnings-for-policy-violations-and-mutations.

PR #64, opened 2024-12-10
doc(KDP): add policy status KDP

KhaledEmaraDev:


proposals/policy_status.md
```diff
@@ -0,0 +1,183 @@
...
+- **What is the expected outcome?** Clear, actionable status information for each policy will enable users to quickly resolve issues, prevent policy deployment failures, and ensure that policies are enforced as intended. The modification to the `kubectl` output will provide a quick overview of policy readiness at a glance.
+
+# Proposal
+
+[proposal]: #proposal
+
+This proposal introduces new status conditions to the Kyverno Policy CRD. These conditions will track the following:
+
+- `WebhookConfigured`: Indicates whether the webhooks required for the policy are set up correctly. Relevant only for Admission controllers.
+- `CachePresence`: Indicates whether the policy has been successfully loaded into Kyverno's internal cache.
```
JimBugwadia, line 65:
```suggestion
- `PolicyCached`: Indicates whether the policy has been successfully loaded into Kyverno's internal cache.
```

proposals/policy_status.md
```diff
@@ -0,0 +1,183 @@
...
+- `RBACPermissions`: Indicates whether Kyverno has the necessary RBAC permissions to enforce the policy.
```
JimBugwadia, line 66:
```suggestion
- `RBACPermissionsGranted`: Indicates whether Kyverno has the necessary RBAC permissions to enforce the policy.
```

proposals/policy_status.md
```diff
@@ -0,0 +1,183 @@
...
+
+Each condition will have a status of `True` or `False`. If a condition's type is `NotReady` and the status is `False`, the policy will fail to apply unless explicitly configured otherwise through `validationFailureActionOverrides`. A condition may not be present if it's not relevant to the policy.
```
JimBugwadia, line 68: > If a condition's type is `NotReady` and

What is this referring to....is there a `type`?

proposals/policy_status.md
```diff
@@ -0,0 +1,183 @@
...
+- **What use cases does it support?**
+  - Quickly identifying the root cause of policy failures (e.g., webhook misconfiguration, missing permissions).
+  - Automating monitoring and alerting based on policy readiness.
+  - Improving the overall reliability of policy enforcement.
+  - Gaining a better understanding of Kyverno's operational state.
+- **What is the expected outcome?** Clear, actionable status information for each policy will enable users to quickly resolve issues, prevent policy deployment failures, and ensure that policies are enforced as intended. The modification to the `kubectl` output will provide a quick overview of policy readiness at a glance.
+
+# Proposal
+
+[proposal]: #proposal
```
JimBugwadia, line 60: I recommend we keep the `status` field with values `Ready` and `NotReady` as a summary state. We can set `status: Ready` when all conditions pass.

JimBugwadia (changes requested):

proposals/policy_status.md
KhaledEmaraDev: So, first I think it's deprecated. I will try to get a reference to this. Second, for future proofing I was thinking that different components might interpret conditions differently. So for instance, the background controller might not care about a policy being configured in the webhook or not. Ready here should be false for Admission but true for Background.

I think it's better if we leave each component to interpret the readiness on its own. But, I'm also fine with this.

proposals/policy_status.md
KhaledEmaraDev: Good catch. This is a typo. The types are what we defined above.

PR #68, opened 2025-01-14
create target config proposal

aerosouund:
Upstream issue: https://github.com/kyverno/policy-reporter/issues/436

### DIFF
--- a/proposals/target_config.md
+++ b/proposals/target_config.md
@@ -1,0 +1,158 @@
+# Meta
+[meta]: #meta
+- Name: (Target Configuration CRD for policy reporter)
+- Start Date: (2025-01-14)
+- Author(s): (aerosouund)
+
+# Table of Contents
+[table-of-contents]: #table-of-contents
+- [Meta](#meta)
+- [Table of Contents](#table-of-contents)
+- [Overview](#overview)
+- [Proposal](#proposal)
+- [Implementation](#implementation)
+- [Unresolved Questions](#unresolved-questions)
+
+# Overview
+
+The [policy reporter](https://github.com/kyverno/policy-reporter) takes in configuration for targets that it will send policy reports to through a flag `--config` which points to a file that can have a structure which is documented [here](https://kyverno.github.io/policy-reporter/core/targets). This document aims to define a CRD that will be used to provide the same functionality as this config file. giving administrators the ability to manage targets with this CRD.
+
+A single CRD equals exactly one target. For example if you want to send to multiple slack channels, you need a CRD per channel
+
+
+# Proposal
+
+The CRD is proposed to have the following structure. The resource is non namespaced, and by default applies to all existing namespaces.
+
+```yaml
+apiVersion: policyreporter.io/v1alpha1
+kind: TargetConfig
+metadata:
+  name: test
+status:
+ sendStatus:
+ - reportName: 179db659-a29f-42c5-a9c2-4106d01ea339
+   succeeded: true
+   messagee: Send OK
+ - reportName: 179db759-a22t-4hf5-al12-4106d01hdbn8
+   succeded: false
+   message: Got response 403
+ conditions:
+  - type: InSync
+    status: "True"
+    reason: SendReportSuccess
+    message: "Sent with report with status 200"
+    lastTransitionTime: "2025-01-07T14:00:00Z"
+spec:
+  targetType: string
+  config: 
+    filter:
+      namespaces:
+        include: ["team-a-*"]
+      priorities:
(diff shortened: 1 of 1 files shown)


proposals/target_config.md
```diff
@@ -0,0 +1,158 @@
...
+
+S3:
+
+```yaml
+apiVersion: policyreporter.io/v1alpha1
+kind: TargetConfig
+metadata:
+  name: s3-target
+spec:
+  targetType: s3
```
fjogeleit, line 81: you should consider to use different keys per target. `config` would be to generic, makes it hard to describe which field is required for which target or even available for a given target type. Makes also the use of the explain command for a given target type easier.

proposals/target_config.md
```diff
@@ -0,0 +1,158 @@
...
+The CRD is proposed to have the following structure. The resource is non namespaced, and by default applies to all existing namespaces.
+
+```yaml
+apiVersion: policyreporter.io/v1alpha1
+kind: TargetConfig
+metadata:
+  name: test
+status:
+ sendStatus:
+ - reportName: 179db659-a29f-42c5-a9c2-4106d01ea339
```
fjogeleit, line 34: its not always the case that a send relates to a single report. This is only the case if the target supports batch send and the policy report only relates to a single resource. Keep in mind that policy reporter also supports different tools which has different approaches.

fjogeleit: You could consider to create Kubernetes Events instead.

proposals/target_config.md
```diff
@@ -0,0 +1,158 @@
...
+    bucket: 'policy-reporter'
+    accessKeyId: 'redacted'
+    secretAccessKey: 'redacted'
+    minimumSeverity: warning
+    skipExistingOnStartup: true
+```
+
+Slack (Inline):
+
+```yaml
```
fjogeleit, line 94:
```suggestion
apiVersion: policyreporter.io/v1alpha1
kind: TargetConfig
metadata:
  name: slack-target
spec:
  slack:
    webhook: "https://hooks.slack.com/services/456..."
  filter:
    namespaces:
      include: ["team-b-*"]
  minimumSeverity: warning
  skipExistingOnStartup: true
```

proposals/target_config.md
aerosouund: I was thinking of this as well, the benefit of doing this was simplicity in my opinion.
if we want to be more descriptive we can use `s3Config`, `slackConfig`.. etc. and the spec can contain any single one of those keys

proposals/target_config.md
```diff
@@ -0,0 +1,158 @@
...
+  name: slack-target
+spec:
+  targetType: slack-inline
+  config:
+    minimumSeverity: warning
+    skipExistingOnStartup: true
+    webhook: "https://hooks.slack.com/services/456..."
+    filter:
+      namespaces:
+        include: ["team-b-*"]
```
fjogeleit, line 107:
```suggestion
```suggestion
apiVersion: policyreporter.io/v1alpha1
kind: TargetConfig
metadata:
  name: slack-target
spec:
  slack:
    webhook: "https://hooks.slack.com/services/456..."
  filter:
    namespaces:
      include: ["team-b-*"]
  minimumSeverity: warning
  skipExistingOnStartup: true
```
```

File: proposals/target_config.md (hunk 1 above)
Comment (fjogeleit) [reply proposals/target_config.md:81]: Do you mean different CRDs per Target? Might be to much. As in my example you can use `spec.s3`, `spec.slack` etc. to group target specfic configuration. You can still keep global configurations like `filter` or `secretRef` directly under `spec`.

Comment (aerosouund) [reply proposals/target_config.md:81]: No, same CRD. but its spec is defined as
```yaml
spec:
  # general keys
  s3: ...  # you can only have one of these
  slack: ...
```
and we can include the general keys separate from each individual config, i suggested this in case at any point in the future we created a struct that an array of those individual configs. in such a case each array item will be its own full config. what do you think ?

proposals/target_config.md
fjogeleit: test

fjogeleit: I mean you create an Kubernetes Event resource when a notification was send or failed instead of updating the status for the target.

proposals/target_config.md
fjogeleit: looks better to me.

> in such a case each array item will be its own full config. what do you think ?

could be a future feature but nothing I would prioritize because it makes many things just more complicated without much benefit. We should keep it simple for now.

realshuting (approved):

PR #74, opened 2025-05-01
feat: adding new a trace command

kushal9897:
we can also make it as a flag

### DIFF
--- a/proposals/trace.md
+++ b/proposals/trace.md
@@ -1,0 +1,134 @@
+# Kyverno Design Proposal (KDP): `kyverno trace` Command
+
+##  What is the `trace` Command?
+
+The `trace` command helps users understand **why a Kyverno policy worked or didn’t work** on a given resource.
+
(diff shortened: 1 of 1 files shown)


proposals/trace.md
```diff
@@ -0,0 +1,134 @@
...
+```plaintext
+ Image: nginx:1.23
+ Attestor 1 (user@example.com)
+ CEL: request.userInfo.groups AnyIn ["prod"] → true
+ Signature verified → Attestor passed
+```
+
+---
+
+##  Why It Helps With Secrets
```
JimBugwadia, line 104: Not sure why we have this section....are Secrets different than any other API call?

JimBugwadia (commented):
Great idea, thanks for submitting! 

Wondering is there is a way to run a `trace` as a `debug` container on an admission controller. This could also be a way to produce a policy evaluation / decision log to an external system.

PR #75, opened 2025-05-07
New Policy Type - DeletingPolicy

fjogeleit:


proposals/cleanup_policy.md
```diff
@@ -0,0 +1,65 @@
...
+    - expression: "variables.name == 'example'"
+  deletionPropagationPolicy: Foreground
+  schedule: "*/2 * * * *"
+```
+
+The new API uses matchConstraints similar to the native VAP / MAP and new kyverno policy types VPOL and IVPOL. The conditions now use the MatchCondition API and are based on CEL expressions to targeting resources.
+
+* CleanupPolicies will support variables. They can be used in conditions
+* Schedule defines a valid cron expression to define when the cleanup policy will be executed
+* DeletionPropagationPolicy defines how resources will be deleted (Foreground, Background, Orphan)
```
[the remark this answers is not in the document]
fjogeleit: Its a port from the current cleanup policy to the new one, so I did not describe it in detail. The basic summary is:

The Deletion Propagation Policy dictates how Kubernetes handles the deletion of dependent objects when a parent object is deleted. It defines the order in which objects are deleted, either with the parent first (Background) or the children first (Foreground). If the policy is set to "Orphan", owner references are ignored, and dependent objects are deleted immediately regardless of the parent's state.

proposals/cleanup_policy.md
```diff
@@ -0,0 +1,65 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      resources:   ["pods"]
+  variables:
+    - name: name
+      expressions: >
+        target.metadata.name
+  conditions:
```
realshuting, line 53: Should we re-name it to `matchConditions` to be consistent with vpol/ivpol?

proposals/cleanup_policy.md
```diff
@@ -0,0 +1,65 @@
...
+    - expression: "variables.name == 'example'"
+  deletionPropagationPolicy: Foreground
+  schedule: "*/2 * * * *"
+```
+
+The new API uses matchConstraints similar to the native VAP / MAP and new kyverno policy types VPOL and IVPOL. The conditions now use the MatchCondition API and are based on CEL expressions to targeting resources.
+
+* CleanupPolicies will support variables. They can be used in conditions
```
realshuting, line 61: Do we want to align it with vpol/ivpol to not support variables in conditions?

proposals/cleanup_policy.md
fjogeleit: no because its different from matchConditions of other policy types as it supports our custom libs and variables what other matchConditions do not allow.

proposals/cleanup_policy.md
fjogeleit: for CleanupPolicies the conditions are the core functionality. It does not have any other CEL logic. So we should provide the most features in the conditions section

JimBugwadia (changes requested):

proposals/cleanup_policy.md
```diff
@@ -0,0 +1,65 @@
...
+    - MatchConstraints for resource selection
+    - Conditions which using the MatchCondition API to define CEL based expressions for resource targeting
+
+## CleanupPolicy API
+
+This example showcases the conversion of the current Cleanup Policy API to the new CEL based approach.
+
+```yaml
+apiVersion: policies.kyverno.io/v1alpha1
+kind: CleanupPolicy
```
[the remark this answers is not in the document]
fjogeleit: renamed

realshuting (commented):
Forgot to publish the comment!

proposals/cleanup_policy.md
```diff
@@ -0,0 +1,65 @@
...
+    - expression: "variables.name == 'example'"
+  deletionPropagationPolicy: Foreground
+  schedule: "*/2 * * * *"
+```
+
+The new API uses matchConstraints similar to the native VAP / MAP and new kyverno policy types VPOL and IVPOL. The conditions now use the MatchCondition API and are based on CEL expressions to targeting resources.
+
+* DeletingPolicies will support variables. They can be used in conditions
+* Schedule defines a valid cron expression to define when the deleting policy will be executed
+* DeletionPropagationPolicy defines how resources will be deleted (Foreground, Background, Orphan)
```
realshuting, line 63: Do we want to support annotation based cleanup policy with deletionPropagationPolicy 
 as described [here](https://kyverno.io/docs/policy-types/cleanup-policy/#ttl-based-cleanup-example-with-deletionpropagationpolicy) ?

realshuting (approved):

MariamFahmy98 (approved):

JimBugwadia (approved):

PR #77, opened 2025-07-30
kdp: add fine-grained exceptions design proposal

MariamFahmy98:
This PR adds a design proposal for fine-grained exceptions for the new policy types.

### DIFF
--- a/proposals/cel_finegrained_exceptions.md
+++ b/proposals/cel_finegrained_exceptions.md
@@ -18,7 +18,7 @@ This proposal outlines the design for a fine-grained exception for Kyverno's CEL
 
 The goals of this proposal are to:
 
-1. **Excluding Specific Images:** A straightforward method to exempt resources from policy enforcement based on their container image reference.
+1. **Excluding Specific Images:** A straightforward method to exempt resources from policy enforcement based on their container image reference. This is important because other identifiers like resource names or labels can be easily spoofed by a user to bypass a policy, an image reference is a strong, trustworthy identifier that points directly to the specific software code being executed. By using the image references in exceptions, we ensure that only the intended, verified software can bypass a rule, making the entire exception mechanism more secure and reliable.
 
 2. **Allowing Specific Values:** A powerful and generic mechanism to allow a resource to contain a specific configuration value (e.g., a hostPath volume, a privileged security context) that a policy would normally deny.
 
@@ -38,8 +38,8 @@ The main idea is to use the existing `PolicyException` resource to gather except
 
 If one or more PolicyException resources match an incoming resource for a given policy, Kyverno will aggregate their specifications and populate two new context variables:
 
-1. `kyverno.excludedImages`: A list of image strings to be exempted.
-2. `kyverno.allowedValues`: A structured map containing permissible values for specific fields in the resource.
+1. `exceptions.allowedImages`: A list of image strings to be exempted.
+2. `exceptions.allowedValues`: A structured map containing permissible values for specific fields in the resource.
 
 If no exception applies, these variables will be empty (empty list and empty map, respectively), ensuring the policy functions normally without any modification.
 
@@ -151,7 +151,7 @@ spec:
   - nginx:latest
 ```
 
-In order to use the excluded images from the exception in the policy, Kyverno will introduce a new CEL object `kyverno.excludedImages` that will be available in the CEL expressions. This object will contain the list of images specified in the PolicyException resource. If no exception exists, `kyverno.excludedImages` is an empty list, and the policy functions normally.
+In order to use the excluded images from the exception in the policy, Kyverno will introduce a new CEL object `exceptions.allowedImages` that will be available in the CEL expressions. This object will contain the list of images specified in the PolicyException resource. If no exception exists, `exceptions.allowedImages` is an empty list, and the policy functions normally.
 
 The policy can then be modified to use this new object in the validation expressions as follows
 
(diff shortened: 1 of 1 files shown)


proposals/cel_finegrained_exceptions.md
```diff
@@ -0,0 +1,320 @@
...
+
+Writing complex CEL logic within the policy itself to handle these one-off cases is hard to manage, and violates the principle of separating policy definition from exceptions. This proposal introduces a dedicated, structured way to manage these fine-grained exceptions.
+
+# Proposal
+
+The main idea is to use the existing `PolicyException` resource to gather exception data and inject it into the CEL evaluation context. This makes the exception data available to the ValidatingPolicy expression through a new kyverno CEL variable.
+
+If one or more PolicyException resources match an incoming resource for a given policy, Kyverno will aggregate their specifications and populate two new context variables:
+
+1. `kyverno.excludedImages`: A list of image strings to be exempted.
```
JimBugwadia: Should we call this `kyverno.allowedImages` to be consistent with `kyverno.allowedValues`?

proposals/cel_finegrained_exceptions.md
```diff
@@ -0,0 +1,320 @@
...
+- [Implementation](#implementation)
+
+# Overview
+[overview]: #overview
+
+This proposal outlines the design for a fine-grained exception for Kyverno's CEL-based ValidatingPolicy.
+
+The goals of this proposal are to:
+
+1. **Excluding Specific Images:** A straightforward method to exempt resources from policy enforcement based on their container image names.
```
JimBugwadia:
```suggestion
1. **Excluding Specific Images:** A straightforward method to exempt resources from policy enforcement based on their container image reference.
```

JimBugwadia: Would be good to explain why this is important.

Container names, resource names, etc. can be spoofed. An image reference cannot.

proposals/cel_finegrained_exceptions.md
```diff
@@ -0,0 +1,320 @@
...
+  policyRefs:
+    - name: disallow-privilege-escalation
+      kind: ValidatingPolicy
+  # list of image names to exclude
+  images:
+  - busybox:latest
+  - nginx:latest
+```
+
+In order to use the excluded images from the exception in the policy, Kyverno will introduce a new CEL object `kyverno.excludedImages` that will be available in the CEL expressions. This object will contain the list of images specified in the PolicyException resource. If no exception exists, `kyverno.excludedImages` is an empty list, and the policy functions normally.
```
JimBugwadia: Would it be more readable to be `exceptions.allowedImages` and `exceptions.allowedValue`?

proposals/cel_finegrained_exceptions.md
JimBugwadia: See below....maybe use `exceptions.*` instead of `kyverno.*`?

proposals/cel_finegrained_exceptions.md
```diff
@@ -0,0 +1,320 @@
...
+
+The policy can then be modified to use this new object in the validation expressions as follows
```
JimBugwadia, line 155: Who modifies the policy?

Does Kyverno automatically do this, or does the user need to modify it? 

If Kyverno can automatically modify (preferred), will we keep the modified policy in the `status` like with autogen? We could potentially have a `status.exceptions.images` and `status.exceptions.values` or simply have the modified CEL in the status?

For other exceptions, we automatically filter out during the `match`. 

How do we automatically do this

JimBugwadia (commented):
A few questions on where and how the modified policy CEL will be managed...

proposals/cel_finegrained_exceptions.md
MariamFahmy98: It would be challenging for Kyverno to automatically modify the policy itself because CEL expressions lack a fixed structure. We cannot parse and insert exception logic into arbitrary CEL conditions without risking altering the user's intent.

For example, consider this expression:
```
expression: (
  has(container.securityContext) &&
  has(container.securityContext.seccompProfile) &&
  container.securityContext.seccompProfile.type == "RuntimeDefault"
) || container.name.startsWith("test-")
```

Where should an exception be added? Wrapping the whole expression, inserting inside the `&&` group, or changing the `||` clause could all modify the policy's behavior.

Another example:
```
expression: object.spec.containers.all(c, c.resources.limits.cpu != "0")
```
Adding an exception here requires rewriting the `.all(...)` logic to conditionally skip containers based on the image, like:
```
object.spec.containers.all(c,
  kyverno.excludedImages.exists(img, c.image.startsWith(img)) ||
  c.resources.limits.cpu != "0"
)
```
This might be correct in logic, but it's not always easy to determine if it applies to all containers, any, or just the first.

If the loop has multiple conditions joined with `&&`, should the exception apply to all or just part?
```
expression: container.securityContext.runAsNonRoot == true && container.name != "init"
```
Here, adding an `||` clause at the top level completely changes the logic. The original intent was to validate both conditions; changing it to:
```
expression: kyverno.excludedImages.exists(i, container.image.startsWith(i)) ||
  (container.securityContext.runAsNonRoot == true && container.name != "init")
```
means that any excluded image bypasses all validations, not only `runAsNonRoot`.

In my opinion, it’s safer to provide exception objects like `kyverno.excludedImages` and `kyverno.allowedValues` for users to explicitly use in their expressions according to their needs/use cases.

proposals/cel_finegrained_exceptions.md
MariamFahmy98: Yes, I agree!

proposals/cel_finegrained_exceptions.md
realshuting: Can we apply exceptions to resources and then apply polices?

PR #78, opened 2025-09-01
add KDP for HTTP authorization

aerosouund:


proposals/http_authorization.md
```diff
@@ -0,0 +1,435 @@
...
+```yaml
+apiVersion: envoy.kyverno.io/v1alpha1
+kind: ValidatingPolicy
+metadata:
+  name: require-foo-header
+spec:
+  evaluation:
+    mode: HTTP
+  validations:
+  - expression: >
```
fjogeleit, line 147: As we reusing the same API (definition) an expression has to return a boolean value, right?

aerosouund: If you're referring to the fact that CEL expressions need to return a bool then no. It has to return a `http.response` type or null. To avoid the conflict, the envoy plugin (or whatever will handle HTTP authorization) will not handle policies that don't have `evaluation.mode: HTTP`. Likewise, kyverno won't handle policies that have this enabled

fjogeleit: and in which sense is it a validation when you returning/building a response object rather then do a validation? This also means that you can not reuse the existing VPOL engine?

aerosouund: Yeah exactly, there's another compiler implementation in the envoy plugin built for those use cases and thats what i am targeting. The plugin also recently added support for VPOL. It is validation in the sense that i am returning a response to a after (validating) a request

PR #89, opened 2026-05-06
[Feature-Proposal] Per-Policy Background Scan Interval 

VarshaUN:
As discussed in yesterday's Kyverno maintainers meeting , I am drafting a proposal for Per-Policy Background Scan Interval 

Signed-off-by : Varsha U N <varshaun58@gmail.com>

cc : @realshuting , @JimBugwadia , @aerosouund


proposals/per_policy_background_scan_inter.md
```diff
@@ -0,0 +1,83 @@
...
+Add a new optional field in the Policy/ClusterPolicy spec:
+
+```yaml
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: require-pod-labels
+spec:
+  background: true
+  backgroundScanInterval: 15m     # <-- new field
```
aerosouund, line 39: @VarshaUN 
i would recommend this setting being controlled through a label or annotation. not a field. given that the old policies are deprecated + it should be a solution generic to both the cluster policy and the cel based policies

JimBugwadia: Hi @aerosouund - I realize we added an annotation to [skip reports](https://github.com/kyverno/kyverno/issues/13886), as we wanted that behavior for VAPs and MAPs as well. 

Why not make this part of the policy API for the Kyverno types?  I prefer keeping the configuration in one place, so its easy for users to manage and track behaviors. I recall this was one of the reasons why we added `background` as a object.

```sh
k --context kind-kind explain vpol.spec.evaluation.background
GROUP:      policies.kyverno.io
KIND:       ValidatingPolicy
VERSION:    v1

FIELD: background <Object>


DESCRIPTION:
    Background  controls policy evaluation during background scan.

FIELDS:
  enabled       <boolean>
    Enabled controls if rules are applied to existing resources during a
    background scan.
    Optional. Default value is "true". The value must be set to "false" if the
    policy rule
    uses variables that are only available in the admission review request (e.g.
    user name).
```
    
 We can discuss separately if this feature is also required for VAPs / MAPs, or if the default intervals are sufficient for them.

aerosouund: @JimBugwadia 
yeah it can be more useful to have this custom interval on the new policy apis. and as an annotation on the old policies.
but don't VAPs/MAPs also have the ability to have this custom reporting scan interval ? if it was through an annotation, then the pattern can be extended to them as well. and if both the old policies and the kubernetes policies will have this configuration through an annotation then maybe its better to make it through an annotation for all. wdyt ?

realshuting: Agreed with Ammar, we need to decide whether we want to apply this VAP/MAP. IMHO we should cover those two types as well, so does how they work in other Kyverno features.

JimBugwadia: @aerosouund @realshuting  - we can allow annotations, but should also have this in the API, so its well documented for users. We can mention in the API docs that the annotation can also be used.

Annotations should not be overused, as the UX is not as great as having a declared API.

